PR #5, opened 2015-06-05
Fixed 469299, 469302, 469305 and addressed committer guidelines 

SmithRWORNL:
Reissuing this pull request with fixes for  469299, 469302, and 469305. I have also fixed my commit log  to reflect the requirements in the Eclipse guideline document for contributing via GitHub.


jayjaybillings:
I am waiting on some feedback from the Eclipse EMO before I merge this since it is not passing the validation step.

jayjaybillings:
I am merging this pull request since we have CLAs on file for both r8s@ornl.gov and smithrw@ornl.gov, which can be validated through the project management tools. I think this page is only updated during the initial commit or pull request, so it has not rescanned the commits after we fixed the CLAs.

PR #10, opened 2015-06-19
Implemented MaterialStackTester and fixed build bug

KasperGam:
MaterialTester was failing due to improper implementation of the equals method for MaterialStack. It did not override the Object.equals method and thus would not compare properly inside the TreeMap (or any container for that matter). Fixed bug 470600.


jayjaybillings:
Passed verification and is a simple bug fix. Accepting.

PR #12, opened 2015-06-24
Merged eclipse/master. Flushed out last implementation for the Reflectivity Model.

KasperGam:
Still needs bug fixes like crazy! Mostly UI reloading and update bugs. The model does work but is not polished yet.


jayjaybillings:
I am rejecting this request because the changes do not seem correctly counted.

PR #16, opened 2015-06-26
[470835]Applied Short Circuit Logic in ToolItemMenuListener

SmithRWORNL:
An instance of ToolItemMenuListener using eager logic instead of short
circuit logic in comparision has been changed. The conditional statement
will now only evaulate the secondary statement if the first returned
false.

Bug: 470835 https://bugs.eclipse.org/bugs/show_bug.cgi?id=470835


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=470835

jayjaybillings:
Accepted.

PR #18, opened 2015-06-26
[470844]MeshAppState floating point equality checking repaired

SmithRWORNL:
MeshAppState no longer directly tests for the equality of two floating
point numbers. It instead properly takes the absolute value of their
difference and compares it to a small delta value.

Bug: 470844 https://bugs.eclipse.org/bugs/show_bug.cgi?id=470844


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=470844

jayjaybillings:
Accepted.

PR #19, opened 2015-06-26
470915]Applied Short Circuit Logic MasterDetailsComponent

SmithRWORNL:
MasterDetailsComponent now uses short circuit comparisions instead of
eager ones. Thus, it will now only evaulate expressions in an AND until
one evaluates to false, instead of evaluating all of them.

Bug: 470915 https://bugs.eclipse.org/bugs/show_bug.cgi?id=470915


jayjaybillings:
Accepted.

jayjaybillings:
Accepted.

PR #22, opened 2015-06-30
Added scripting from Eclipse EASE.

KasperGam:
ICE now can be accessed from a script using the modules provided by EASE. Use of either Jython or Javascript is supported, although Jython is not pre-installed. 

Fixed bug 471370.


jayjaybillings:
Pulling since it addresses a feature request and is small.

PR #26, opened 2015-07-10
[472380]Repaired SWTBot tests

SmithRWORNL:
Adds proper dependencies to SWTBot test bundles, allowing them to create
Eclipse instances when launched from Maven to run tests in. It also
changes PlotEditorTester to use a fake IVizServiceFactory, since Maven
does not iniatalize and consume all OSGi services needed for the test.

Bug: 472380 https://bugs.eclipse.org/bugs/show_bug.cgi?id=472380


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=472380

PR #27, opened 2015-07-17
Smith rwornl/refactor

SmithRWORNL:
Splits the GeometryComponent into GeometryComponent and Geometry, and moves the Geometry class to a new viz service bundle.


jayjaybillings:
This is less than 1000 lines and address active bugs, so pulling. Thank you!

PR #29, opened 2015-07-24
Smith rwornl/refactor

SmithRWORNL:
Moved the geometry code into a new geometry service bundle.


amccaskey:
This has been approved by the CQ process, were accepting it under parallel IP.

PR #37, opened 2015-09-26
Jay/project support

menghanli:
Bug 478474 was fixed to transform IMaterialsDatabase OSGi services into Extension Point and Extension Points


jayjaybillings:
It addresses an open bug, passes the IP checks and is less than 1000 lines, so I will try to merge it manually and get back to you.

PR #39, opened 2015-10-07
Transform OSGi services in Reactor into ExtensionPoints

menghanli:
Bug 479178 is fixed to transfrom OSGi services in Reactor into ExtensionPoints and Extensions.

Signed-off-by: menghan <menghanli723@gmail.com>


jayjaybillings:
Addresses a bug, is less than 1000 lines and passes checks. Accepted.

PR #40, opened 2015-10-08
Initial public commit with latest geometry editor work.

enleeten:
Change-Id: I516583ae88727bedd318bd410d2728f80e72c9b0
Signed-off-by: Tony McCrary <tmccrary@l33tlabs.com>


amccaskey:
Hey Tony, 

Just tried to pull this down and build. It fails because it can't satisfy dependencies to some of the new geometry bundles - apparently because the pom version numbers are different from the Manifest versions.

amccaskey:
Also, ICE development is not taking place on the "next" branch. Could you please merge next into your branch and then submit that back to next?

enleeten:
I will get these issues update and issue a new request.

PR #45, opened 2015-10-21
Moved the XMLPersistenceProvider.getJAXBProviders()operation.

menghanli:
I moved this operation to be a static interface operation on
IJAXBClassProvider, which will remove the requirements for the
XMLPersistenceProvider

Signed-off-by: Menghan Li <menghanli723@gmail.com>


jayjaybillings:
Do you have a number for a bug report that this is tied to?

I will accept the provider changes, but we don't need the Debug makefiles. I will try to merge it now.

jayjaybillings:
Bug 479184 covers this. Nick created it awhile back and I'll tag this work to that.

jayjaybillings:
Seeing as all checks pass, that the code is less than 1000 lines and that I have been able to merge this locally, I accept this pull request. Thank you!

PR #51, opened 2015-11-15
Bug 481365

menghanli:
Added extension point for resource page providers.
ICEFormEditor now uses IResourcePageProvider to find all registered
providers, and then uses the default provider to get the pages.

Signed-off-by: Menghan Li <menghanli723@gmail.com>


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=481365

jayjaybillings:
Since this passes the IP checks, addresses a bug and is less than 1000 lines, I am accepting it. Thank you!

PR #52, opened 2015-11-18
Bug 481365 BasicComponentProvider and generic extension helper

nickstanish:
- Created a generic extension helper which will reduce repetitive code of finding extensions to an id and then casting the implementations.
- Added that extension helper to the existing providers
- Created BasicComponentProvider interface and extension point as well as default extension
- Fixed IResourcePageProvider extension point (was using "Class" instead of "class")

### DIFF
--- a/org.eclipse.ice.client.widgets/plugin.xml
+++ b/org.eclipse.ice.client.widgets/plugin.xml
@@ -402,7 +402,8 @@
        name="Default Resource Page Provider"
        point="org.eclipse.ice.client.widgets.resourcePageProvider">
     <Implementation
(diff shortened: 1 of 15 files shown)


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=481365

jayjaybillings:
Nick asked me to wait another hour or so before accepting this because he has one more small thing that he is fixing.

nickstanish:
Ok @jayjaybillings it's ready

jayjaybillings:
This addresses a bug and adds some awesome capabilities, passes the validation checks and is less than 1000 lines, so I am accepting it. Thanks very much @nickstanish!

PR #55, opened 2015-12-02
Bug 482456

wo-amlangwang:
Added extension point for Geometry page providers.
ICEFormEditor now uses IGeometryPageProvide to find all registered
providers, and then uses the default provider to get the pages.

Signed-off-by: Fangzhou Lin <adamlangwang@gmail.com>


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=482456

jayjaybillings:
Great work Fangzhou! The only problem is that you are missing documentation. You need to add copyright information (using your identifying information where appropriate) at the top of each new file and you need to put documentation on each method.

Check out

https://github.com/eclipse/ice/blob/jay/UIRefactor/org.eclipse.ice.client.widgets/src/org/eclipse/ice/client/widgets/providers/DefaultResourcePageProvider.java

and 

https://github.com/eclipse/ice/blob/jay/UIRefactor/org.eclipse.ice.client.widgets/src/org/eclipse/ice/client/widgets/providers/IPageFactory.java

for examples.

Note that you can let Eclipse generate the documentation for overridden methods by typing "/*" and then hitting enter above the @Override statement.

jayjaybillings:
Looks great. Accepted since it is less than 1000 lines, passes all checks and addresses a bug. Thanks very much for your contribution!

Issue #57, opened 2015-12-03
API compilation issues with the latest version of VisIt


jayjaybillings:
ICE's visualization bundles do not compile with the latest version of the visit_java_client written by @HarinarayanKrishnan.

jayjaybillings:
Fixed in Hari's latest commit. Thanks for the contribution!

PR #59, opened 2015-12-03
Fixing Visualization Perspective view to work with new visit_java_cli…

HarinarayanKrishnan:
Fixing Visualization Perspective view to work with new visit_java_client interface. For bug #57

Signed-off-by: Hari Krishnan <harinarayan.krishnan@gmail.com>


jayjaybillings:
I'm rejecting this because the validation issues still exist.

Issue #128, opened 2016-02-25
Move HIPP builds of ICE into downloads directory


jayjaybillings:
We need to configure out HIPP to move our binaries into our downloads directory after a build. @jarrah42 will take this.

erwindl0:
Will you also publish a p2 repository from your build?
Is this feasible using eclipse storage, and to make the result 
accessible via downloads and direct URL?

if so, I will be very interested to see how to get that done, as we 
would need that for Triquetrum as well...

jayjaybillings:
@erwindl0 yes, we need to do that too. Sorry, I should have specified that better. 

We currently publish our binaries and our p2 repo on https://eclipseice.ornl.gov/downloads/ice.

erwindl0:
@jayjaybillings great! 

Just to be 100% sure : you will also offer a p2 access from default eclipse CBI resources, NFS shares, download area etc? Or will that part remain on your own hosted server that you mention above?

If you could share the details about poms, p2 metadata, HIPP job config to get that done, that would help us a lot. 

(In fact I need to have a Triquetrum p2 available I think to support the easy integration in ICE's target for our workflow integration trials.)

jayjaybillings:
@erwindl0 Yes, we will share the p2 repo on Eclipse servers.

Check out our org.eclipse.ice.repository/pom.xml for information on how to build the repository. The p2 repo will be put in the target/ directory of the same bundle once it builds. I don't know how the download will be configured on the Eclipse infrastructure yet (@jarrah42 will help us with that). 

You will need this for the ICE integration, yes. If you want, you can send me a zip file of the p2 directory once you get it built and I can upload it to eclipseice.ornl.gov, just to get the ball rolling on the integration.

erwindl0:
@jayjaybillings the p2 repo is already built on our HIPP as well, similar to what you describe.
It's just not reachable yet afterwards in an easy way, and is stuck indeed somewhere in the workspace...

The zip is somewhat big, but I could already share it with you via a drive service out there....
(until I can duplicate @jarrah42 's work ;-) )

jayjaybillings:
Fixed in @jarrah42's last commit.

jarrah42:
I've enabled this for the "ice-next" build. For the record, this is enabled by adding a "production" profile to the build. This is done on hudson by adding "production" to the Profiles field under the "Advanced" tab of "Invoke Maven 3".

PR #133, opened 2016-02-25
Issue #128: Add production profile to copy build result to downloads area

jarrah42:
Adds <ice-version> and <ice-install-dir> properties in the parent pom.xml.


jayjaybillings:
Fixes pressing bugs and passes all checks. Accepted! Thanks for your contribution!

Issue #106, opened 2016-02-17
Setup ICE application signing 


amccaskey:

jayjaybillings:
@jarrah42 will take this.

PR #144, opened 2016-02-27
Issue #106: Set cbi repository explicitly

jarrah42:
Ok, so I admit that the maven build is a bit of a black art... not to mention that the documentation is crap.

Signed-off-by: Greg Watson <g.watson@computer.org>


jayjaybillings:
Accepted. Thanks.

PR #247, opened 2016-07-23
Add shebang so that ICE installer can be executed directly

jonahkichwacoders:
Without the shebang, on Linux this happens:
$ ./Install_ICE.py
from: can't read /var/mail/__future__

Signed-off-by: Jonah Graham <jonah@kichwacoders.com>


jayjaybillings:
Thanks very much! Accepted since it passes all checks and is small.

Issue #299, opened 2016-11-29
ICE XSEDE Tutorial, section 4.4.1 typo: duplicated word 'the'


klaskyhb:
ICE XSEDE Tutorial, section 4.4.1 typo at:

“4.4.1 Opening a Plot Editor
To open a Plot Editor, a ﬁle that uses this editor must ﬁrst be placed in the Project Explorer. This view lists ﬁles imported into ICE. To access the Project Explorer, use **the the** menu bar at the top of the window and navigate to Window → Show View → Project Explorer.”

jayjaybillings:
As I mentioned in #298 this would be a really good place for you to start making pull requests.

klaskyhb:
Yes, I will work in fixing this error on my own and issuing a pull request.

SmithRWORNL:
Fixed by pull request #311

Issue #298, opened 2016-11-29
ICE XSEDE Tutorial, section 1.2.2 provide the location and the name of the executable file referred.


klaskyhb:
On ICE XSEDE Tutorial, section 1.2.2 it says: “To start ICE, double-click on the application executable to open up ICE.” However, no executable file name is mentioned. Documentation should provide the location and the name of the executable file that it is being referred.

jayjaybillings:
@klaskyhb These files are in the doc/ folder of our directory. Would you be willing to try fixing this error on your own and issuing a pull request?  (See CONTRIBUTING.md for more info.)

klaskyhb:
Yes, I will work in fixing this error on my own and issuing a pull request.

SmithRWORNL:
Fixed by pull request #311

PR #311, opened 2016-12-08
Fix of Bugs #298 and #299

klaskyhb:
Requesting to pull fixes to bugs  #298 and #299.


jayjaybillings:
All checks pass and this addresses open bugs, so accepted. Thank you!

PR #360, opened 2017-03-13
Worked with Jay to try to fix EMF/Tomcat issues.

anarak:


jayjaybillings:
Accepted. Thanks! Passes validation and @anarak works at ORNL and falls under our member agreement, even though she isn't a project committer.

PR #361, opened 2017-03-15
Bind OS package field

anarak:
This includes refactored AppsUI.


SmithRWORNL:
Accepting this commit as Anara is an employee of Oak Ridge National Laboratory and is covered under their member committer agreement.

PR #386, opened 2019-08-21
Continued Commands API functionality

osbornjd:
Continued adding functionality for the Commands API. Added some tests and have started filling out LocalCommand code.


osbornjd:
Two most recent commits address issues [387](https://github.com/eclipse/ice/issues/387) and [388](https://github.com/eclipse/ice/issues/388)

Issue #397, opened 2019-08-28
FileHandlerTest in Commands API should create a file and search for it


jayjaybillings:
The FileHandlerTest in the Commands API needs to create a file in its BeforeAll operation instead of testing against a Linux-specific file.

This supports #384.

jayjaybillings:
@osbornjd finished this.

Issue #396, opened 2019-08-28
Fix parameters that need to be marked "final" in Commands API


jayjaybillings:
We need to review the Commands API and make sure that all parameters that should be marked "final" are marked as such.

This supports #384.

jayjaybillings:
@osbornjd finished this.

Issue #395, opened 2019-08-28
CopyFileCommand and MoveFileCommand have wrong inheritance


jayjaybillings:
CopyFileCommand and MoveFileCommand should inherit from Command, not FileHandler.

This supports #384.

jayjaybillings:
@osbornjd finished this.

Issue #394, opened 2019-08-28
Remove calls to System.exit in Commands API


jayjaybillings:
We should throw exceptions instead of calling System.exit(1) in the Commands API. This supports #384.

jayjaybillings:
@osbornjd finished this.

PR #399, opened 2019-08-28
Additional Commands API features

osbornjd:
## Summary
This PR addresses a number of issues raised in the previous [PR391](https://github.com/eclipse/ice/pull/391).

### Issues addressed in this PR
The issues resolved in this PR are [393](https://github.com/eclipse/ice/issues/393), [394](https://github.com/eclipse/ice/issues/394), [395](https://github.com/eclipse/ice/issues/395), [396](https://github.com/eclipse/ice/issues/396), and [397](https://github.com/eclipse/ice/issues/397). 

### Other changes
This PR also fixes a bug in the job monitoring which was identified during testing.

It also implements an issue raised in the previous PR391 and not mentioned in individual issues threads, which is the implementation of the CommandTest function.

## Potentially (remaining) open issue
It is up for discussion whether or not other places need "final" in reference to [issue 396](https://github.com/eclipse/ice/issues/396) - nonetheless the additions I have made are a beginning of (for what is implemented so far) the things that I think need to be static.

### DIFF
--- a/org.eclipse.ice.commands/src/main/java/org/eclipse/ice/commands/Command.java
+++ b/org.eclipse.ice.commands/src/main/java/org/eclipse/ice/commands/Command.java
@@ -111,7 +111,7 @@ public abstract class Command{
 	 * It replaces ${inputFile}, ${installDir} and other keys from the dictionary. The function
 	 * is abstract so that Local and Remote executable names can be handled individually,
 	 * since the remote target file system is not necessarily the same as the local. 
-	 * It should be overridden by the subclasses that require executables.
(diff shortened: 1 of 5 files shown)


osbornjd:
1.  In the same file, why would return RUNNING at line 372 if we get an exception? Would that not indicate it failed?
 - The exception is that the job hasn't finished yet, not necessarily that it is failed. It could just still be running, which is why we go to monitor the job and wait for it to finish. That is why we return RUNNING at this line. If it actually fails, then CommandStatus is marked as such in jobMonitor()

jayjaybillings:
Good. Thanks.

Issue #402, opened 2019-08-29
Make execute() called by user in all instances


osbornjd:
execute() shows up in the constructor of multiple commands, and should be called explicitly by the user.

osbornjd:
issue resolved in above reference

Issue #401, opened 2019-08-29
Check for file existence as early as possible


osbornjd:
LocalFileCopyCommand should not use FileHandler.exists() since FileHandler depends on it. I would argue that the check for file existence should be done long before line 50. Also, it shouldn't do the actual copy in the constructor. Clients should be required to explicitly call execute().

osbornjd:
issue resolved in above reference

Issue #400, opened 2019-08-29
FileHandlerTest needs to delete the files it creates


osbornjd:
FileHandlerTest should delete the file it creates when it is done using AfterAll. This will delete the temporarily created files.

osbornjd:
Issue resolved in above reference

Issue #398, opened 2019-08-28
Need to run Eclipse Formatter on all Commands API files


jayjaybillings:
The Commands API code should be formatted using the standard Eclipse Formatter to match the expectations of the project. Right now most of the block statements have poorly formatted closing/secondary clauses, and there is substantial white space and misalignment.

The Eclipse Formatter can be used on a per file basis, or through the right-click menu and Refactor submenu on the whole project.

This supports #384.

jayjaybillings:
@osbornjd finished this.

PR #403, opened 2019-08-29
FileHandler tests improved

osbornjd:
## Summary
This PR addresses the issues [392](https://github.com/eclipse/ice/issues/392), [394](https://github.com/eclipse/ice/issues/394), [398](https://github.com/eclipse/ice/issues/398), [400](https://github.com/eclipse/ice/issues/400), [401](https://github.com/eclipse/ice/issues/401), and [402](https://github.com/eclipse/ice/issues/402). Most of these are small changes, but a few details about more significant changes are described below. 

### Changes in FileHandler
The main issue resolved is that FileHandler now returns a Command which the user tells to execute, rather than the execution being performed in the constructor of a CopyFile or MoveFile Command.

FileHandlerTest also now creates and destroys temporary test files.

### Logger
A logger was also added to replace the System.out print statements. The logger prints to the console as defined by a very basic default log4j.properties file that was added to the resources directory. This directory is by default included in the classpath so the maven project picks it up and applies it at run time (e.g. in CommandFactoryTest).

### DIFF
--- a/org.eclipse.ice.commands/pom.xml
+++ b/org.eclipse.ice.commands/pom.xml
@@ -11,5 +11,15 @@
 			<artifactId>junit</artifactId>
 			<version>4.4</version>
 		</dependency>
+		<dependency>
+      		<groupId>org.slf4j</groupId>
+     	    <artifactId>slf4j-log4j12</artifactId>
+            <version>1.7.12</version>
+        </dependency>
+        <dependency>
+            <groupId>log4j</groupId>
+            <artifactId>log4j</artifactId>
+           <version>1.2.17</version>
+        </dependency>   
 	</dependencies>
 </project>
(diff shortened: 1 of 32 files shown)


jayjaybillings:
All checks pass. A few notes are provided in my review above. @osbornjd is covered by ORNL's MLA. Accepted. Thanks!

osbornjd:
1. Lines 67 and 113 of FileHandler uses System.out.println().
- Whoops, good catch I thought I had checked that... fixed now.

2. Is there a way to consolidate copy() and move() since they share a lot of the same code? I think if they both started by calling FileHandler.exists() then the shared parts would be eliminated almost entirely.
- Good question, I've been wondering this myself but wondered if we originally intended to keep them separate just for clarity sake. I'll think about it some more and see what I come up with.

Issue #409, opened 2019-09-05
FileHandler.move() places source file in destination directory instead of new file name


jayjaybillings:
Calling FileHandler.move() will move the source file into a destination directory instead of moving the file to a new position. This means that
```java
		FileHandlerFactory factory = new FileHandlerFactory();
		IFileHandler fileHandler = factory.getFileHandler();
		fileHandler.move(filePath, newFilePath);
		return;
```
will not result in /home/jay/testFile1.txt being moved to /home/jay/testFile2.txt. Instead, it becomes /home/jay/testFile2.txt/testFile1.txt with the destination name being used as a directory name.

osbornjd:
Closing this issue as it was fixed and merged some time ago.

PR #408, opened 2019-09-05
Overhaul of FileHandler Design

osbornjd:
## Summary
This PR introduces an overhaul of the FileHandler design and functionality. The overall design is cleaner and largely hidden from the user. There are still some areas of functionality that need to be worked out with remote file transfers, but these can be addressed once the remote command API is finished. 

This PR addresses issue [404](https://github.com/eclipse/ice/issues/404).

### Brief Overview of Design
The FileHandler design now operates as follows:

1. A factory IFileHandlerFactory creates a concrete instance of a Local or Remote FileHandler, which inherits from the abstract class FileHandler. FileHandler implements the interface IFileHandler, which describes the overall layout of the FileHandler design.
- Note that at the moment, IFileHandlerFactory just creates a LocalFileHandler. RemoteFileHandling will be addressed once RemoteCommands have been implemented.
2. The concrete instance of Local/Remote FileHandler then creates a concrete instance of a MoveFileCommand or CopyFileCommand, depending on which file transfer command was called.
3. FileHandler then calls the Command to execute which actually executes the the file transfer. A CommandStatus is passed up the chain to the original call by the user to copy/move a file to confirm that the transfer was completed successfully. 

### Other Changes Unrelated to FileHandler

A few changes were made that were unrelated to the FileHandler

1. The comment structure for the getter/setter functions in CommandConfiguration was adjusted to meet the eclipse style.
2. A test in TestCommandFactory was adjusted so that the working directory can be identified on a user's computer and the test script can then be run without absolute paths being specified.

PR #410, opened 2019-09-05
FileHandler Changed to Using Accessors

osbornjd:
In conjunction with the recommendation from [PR 408](https://github.com/eclipse/ice/pull/408), this PR changes the design of the FileHandler slightly to use default constructors and then set the source and destination files via accessors, rather than setting them in the constructor.

This [commit](https://github.com/eclipse/ice/pull/410/commits/7f4e4f559ed62f3eb633c9b67964fc0fc07de354) also addresses the bug pointed out by @jayjaybillings in issue [409](https://github.com/eclipse/ice/issues/409).

There is also a commit that adds two example files that demonstrate how the commands API works for the user.


jayjaybillings:
@osbornjd can you resolve the merge conflict please? I think you need to pull the latest from ICE and then merge that into your code.

I believe the conflict is that I changed the name of a class and function.

osbornjd:
@jayjaybillings I believe this is ready to be reviewed now.

One comment on the previous commit - I moved the instantiation of the LocalMoveFileCommand (as well as copy) into setMoveConfiguration (or copy) instead of in the constructor of the LocalFileHandler. I think this is much cleaner because it allows the base class to have a single Command member variable which is then instantiated depending on what the user calls.

If the creation of the command is left in the constructor, it requires you to have two member variables CopyFileCommand and MoveFileCommand and basically you need a separate track of code to deal with each of the copy/move logic flows. This is not ideal since all of the logic is nearly identical, so it is highly desired to have a single Command member variable to reduce the amount of code in the classes.

The saving is that in LocalFileHandler we have two (nearly identical) setConfiguration functions, whereas if we have the creation in the constructor of the LocalFileHandler we need to have 4 nearly identical functions.

osbornjd:
I left these blank because the OS is set by default now - I suppose I should have added them back in to the tests for demonstration purposes. 

But adding in || commandConfig.getOS() == null back into Command seems not useful since it is now set by default in the constructor, so that will always pass.

jayjaybillings:
This looks good. It appears to have addressed all of my change requests and then some. All checks pass and @osbornjd is covered under the UT-Battelle MLA.

PR #411, opened 2019-09-11
ConnectionManager and Associated Connections

osbornjd:
I'm going to introduce this PR now since I think it would be useful to review the code associated with connections and get comments on it as a "first step" towards filling out the rest of the remote commands. 

This PR introduces some code that manages remote connections. There are three main classes, ConnectionConfiguration, Connection, and ConnectionManager which is a static factory. ConnectionManager is responsible for opening, closing, and in general managing the various remote connections that one might open. ConnectionConfiguration and Connection are the two classes which contain all of the relevant information necessary for establishing/holding/closing a remote ssh connection.

At the moment, the test for ConnectionManager just tests the ability to successfully open and close an ssh connection with a private (to me) ssh connection. I am planning on making this more general so that tests can be run by anyone without giving away personal ssh credentials.


osbornjd:
Here are some comments to these comments. A lot of these were addressed in commits that are pushed now after filling out the remote command functionality.

> What was your reasoning about having ConnectionConfiguration inherit from UserInfo and UIKeyboardInteractive?

I at first thought this had to be done for JSch password implementation, but have since removed it.

> What's the purpose of ConnectionManager.getPassword()?

The purpose is to get the password from the user at the console prompt so that the password doesn't have to be stored in a string variable, which is inherently unsafe. I think the comment in this function should reflect this as it says "This function gets the password from the user as a prompt" amongst other things.

>  We need some more documentation in RemoteCommand. For example, at line 151, why are we creating a working directory? (We know, but would another developer?)

This PR was originally supposed to just address the connection managing, the remote command code is now filled out and has much more detail. It is included in this PR now in the commits from the previous few weeks.

> ConnectionManagerTest line 58 should assume the file is in user.home so that you can remove the hard coded path.

This has been changed since setting up the dummy ssh account with Caleb. The file with the dummy credentials now lives in `/tmp/`.

>  I think you need to review the design of ConnectionManager test. If the tests have to be executed in a specific order, it could be the case that they are not being tested correctly to begin with. You can hit me up on Slack about this or ask in person on Monday. Also this test appears largely unimplemented at the moment.

I agree with this to some extent, however there is some order that has to be applied for certain cases. For example, it is impossible to test closing a valid connection if a valid open connection does not already exist, so these have to be performed in order. 

Perhaps your comment is that the tests can be expanded to check for exception cases, for example testing that an exception or something is thrown if the user tries to close a connection that doesn't already exist. This I agree with and I have tried to implement some of these cases in the (now more filled out) ConnectionManagerTest.

osbornjd:
## Summary
I have updated this PR with a lot of changes over the course of a few weeks. In addition to filling out most of the ConnectionManager capability, the Remote Command API has now been implemented which enables commands to be executed on a remote host. Tests are performed with a dummy ssh connection, set up [here](https://code.ornl.gov/osbornjd-eclipse/ice-ci). I need to work with Caleb to get this mirrored to the repository, we can probably do that sometime this week.

### ConnectionManager and Connection Hosting
The ConnectionManager functionality and testing is completely filled out. ConnectionConfiguration and Connection are two java classes that are responsible for holding the information regarding how a connection should be made and the actual connection itself. ConnectionManager is responsible for opening, getting, and closing remote connections. It keeps a list of connections as its member variable and adds/removes from the list when needed.

### Remote Commands
The remote commands API has been filled out and tested now. The API uses the JSch package to establish and execute remote commands on another host. At the moment, tests are performed with a dummy ssh account, where the credentials are read in via a text file. Users can establish ssh connections in a similar way, or they can set their username/hostname manually after which they will be prompted at the console for their password. The RemoteCommand has a similar design to the LocalCommand, where there are four major steps - the job environment is setup, then the job is actually processed, the job is monitored to ensure that it finishes successfully, and then the environment is cleaned up and output is logged.

Note - some functionality that was specific to LocalCommands was moved out of the super class Command and into LocalCommand. This makes for a cleaner design and is also more intuitive, since the LocalCommands use the ProcessBuilder API to execute jobs locally and the RemoteCommands use the JSch API to execute jobs remotely. Common functionality (e.g. logging output or setting up the job configuration) is still shared by the super class.

### Remote File Handling
At the moment, only the tests for the remote file handling API are filled out, not the actual source code. This will be introduced in another PR. However, the tests are included in some of these commits.

org.eclipse.ice.commands/src/main/java/org/eclipse/ice/commands/RemoteCommand.java
```diff
@@ -25,7 +36,17 @@
...
+
+	/**
+	 * A file output stream for error messages to be remotely logged to
+	 */
+	FileOutputStream stdErrStream = null;
+
+	/**
+	 * A file output stream for output messages to be remotely logged to
+	 */
+	FileOutputStream stdOutStream = null;
```
SmithRWORNL: Should be private/protected

org.eclipse.ice.commands/src/main/java/org/eclipse/ice/commands/RemoteMoveFileCommand.java
```diff
@@ -27,12 +26,12 @@
 	/**
 	 * The path to the source file which is to be copied
 	 */
-	Path source;
+	String source;
```
SmithRWORNL: Should be private/protected

org.eclipse.ice.commands/src/main/java/org/eclipse/ice/commands/RemoteMoveFileCommand.java
```diff
@@ -27,12 +26,12 @@
...
 
 	/**
 	 * The path of the destination for which the source file will be copied to
 	 */
-	Path destination;
+	String destination;
```
SmithRWORNL: Should be private/protected

osbornjd: Whoops, sorry. I meant to go through and double check this yesterday and it slipped my mind after I talked with Jay. I've gone through and double checked all member variables are private/protected in all of the classes.

SmithRWORNL:
> You need to override equals() on Connection such that the check at line 124 in ConnectionManager can be done with a faster, native search algorithm. We can look at this in more detail on Monday.

This hasn't been done, but I don't see any relevant check in ConnectionManager either, so can you confirm you discussed this with Jay and it's no longer necessary?

Also, it doesn't have to be implemented in this PR necessarily, but I'll note that I don't understand the design behind having exactly one input file per command. In Command line 201, your validity check fails if getInputFile() == null, but shouldn't input file be null in the case that there is no input file, ie when running ls remotely to view remote file system contents? 

Similarly, it would be useful to be able to specify multiple input files in CommandConfiguration. Imagine I'm running a command of the form "qclimax --resolution BASIS400.dat --data BASIS401.dat" where both .dat files are inputs to the program. It is my understanding that all the SHAMAN reduction jobs work like this, where there are multiple files of different kinds specified as command line arguments. My solution would be to make inputFile a list, and the CommandConfiguration would replace input files in the execution by name based on index. ie:

inputFiles = { "BASIS400.dat", "BASIS401.dat"}
executable = "qclimax --resolution ${inputFile0} --data ${inputFile1}"

Also in CommandConfiguration.getExecutableName(), fullCommand is assigned but never used, and should be deleted.

jarrah42:
Here are some things that are going to bite you:

1. JSch is not thread safe. So if you're calling any JSch methods from different threads (and it looks like you are) then you'd better be handling the synchronization or you're going to get corruption/weird behavior.
2. JSch is being phased out of Eclipse in favor of Apache Mina as the former is not supported any more.
3. Some systems do not allow sftp access, so if you're relying on sftp for file manipulation commands then you should at least test if it is available and notify the user if not. Having an alternative you can use as a fallback is also helpful.
4. Many HPC systems (including OLCF) restrict the number of ssh channels to one, so relying on being able to open channels whenever you want is problematic. There are only two ways around this: either support multiple ssh connections per your "connection" which is messy and complicated, or provide a multiplexing protocol which requires an agent on the target machine, also messy and complicated.
5. You will need to support multi-hop connections, where the user connects to an intermediate machine first, and then connects to the target machine from there. Two sets of credentials will need to be provided. Most intermediate systems seem to support tunneling, so you can usually set up a tunnel to the target machine once the connections have been established.

osbornjd:
@SmithRWORNL 
> This hasn't been done, but I don't see any relevant check in ConnectionManager either, so can you confirm you discussed this with Jay and it's no longer necessary?

Yes, Jay and I discussed this. I was previously using an ArrayList and we discussed changing this to a hashmap, which addresses this comment by making it no longer necessary.

> Also, it doesn't have to be implemented in this PR necessarily, but I'll note that I don't understand the design behind having exactly one input file per command. In Command line 201, your validity check fails if getInputFile() == null, but shouldn't input file be null in the case that there is no input file, ie when running ls remotely to view remote file system contents?

That is a fair point, I have a test that tests `ls` on a remote system and had set the inputfile to blank. But it makes more sense to just have no input file, so I've updated this.

> Similarly, it would be useful to be able to specify multiple input files in CommandConfiguration. Imagine I'm running a command of the form "qclimax --resolution BASIS400.dat --data BASIS401.dat" where both .dat files are inputs to the program. It is my understanding that all the SHAMAN reduction jobs work like this, where there are multiple files of different kinds specified as command line arguments. My solution would be to make inputFile a list, and the CommandConfiguration would replace input files in the execution by name based on index. ie:
> 
> inputFiles = { "BASIS400.dat", "BASIS401.dat"}
> executable = "qclimax --resolution ${inputFile0} --data ${inputFile1}"

Very good suggestion, I have edited the code and tests to fit this executable format as well as the possibility if one wants to execute something like "./someExecutable.sh" and then just append all of the input files on the end as arguments.

> Also in CommandConfiguration.getExecutableName(), fullCommand is assigned but never used, and should be deleted.

fullCommand is used for logging the output, e.g. in the function createOutputHeader of CommandConfiguration. This admittedly isn't much, but I don't think it is that bad to leave it as it is. If you disagree I can remove it.

SmithRWORNL:
> fullCommand is used for logging the output, e.g. in the function createOutputHeader of CommandConfiguration. This admittedly isn't much, but I don't think it is that bad to leave it as it is. If you disagree I can remove it.

I had missed the fact that fullCommand was a class variable, so disregard that comment.

The rest of the changes look good. I've got nothing else, so I'll leave it to you and @jarrah42  to sort out how much needs to be done on the connection issues before accepting the request.

org.eclipse.ice.commands/src/main/java/org/eclipse/ice/commands/RemoteMoveFileCommand.java
jayjaybillings: Conversations with me tend to have that effect...

jayjaybillings:
> > You need to override equals() on Connection such that the check at line 124 in ConnectionManager can be done with a faster, native search algorithm. We can look at this in more detail on Monday.
> 
> This hasn't been done, but I don't see any relevant check in ConnectionManager either, so can you confirm you discussed this with Jay and it's no longer necessary?
> 
> Also, it doesn't have to be implemented in this PR necessarily, but I'll note that I don't understand the design behind having exactly one input file per command. In Command line 201, your validity check fails if getInputFile() == null, but shouldn't input file be null in the case that there is no input file, ie when running ls remotely to view remote file system contents?
> 
> Similarly, it would be useful to be able to specify multiple input files in CommandConfiguration. Imagine I'm running a command of the form "qclimax --resolution BASIS400.dat --data BASIS401.dat" where both .dat files are inputs to the program. It is my understanding that all the SHAMAN reduction jobs work like this, where there are multiple files of different kinds specified as command line arguments. My solution would be to make inputFile a list, and the CommandConfiguration would replace input files in the execution by name based on index. ie:
> 
> inputFiles = { "BASIS400.dat", "BASIS401.dat"}
> executable = "qclimax --resolution ${inputFile0} --data ${inputFile1}"
> 
> Also in CommandConfiguration.getExecutableName(), fullCommand is assigned but never used, and should be deleted.

I agree with @SmithRWORNL. The file replacement with ${} is how it works in ICE, although the version you are using as a reference may not have the full replacement algorithm in it.

osbornjd:
@jayjaybillings I think that Robert's suggestion was addressed in this [commit](https://github.com/eclipse/ice/pull/411/commits/8463c1a944a62951d2c45ea9ba4655068048d8e6). You are correct that the reference did not have the full replacement algorithm but I believe I sorted it out.

jayjaybillings:
> Here are some things that are going to bite you:
> 
> 1. JSch is not thread safe. So if you're calling any JSch methods from different threads (and it looks like you are) then you'd better be handling the synchronization or you're going to get corruption/weird behavior.
> 2. JSch is being phased out of Eclipse in favor of Apache Mina as the former is not supported any more.
> 3. Some systems do not allow sftp access, so if you're relying on sftp for file manipulation commands then you should at least test if it is available and notify the user if not. Having an alternative you can use as a fallback is also helpful.
> 4. Many HPC systems (including OLCF) restrict the number of ssh channels to one, so relying on being able to open channels whenever you want is problematic. There are only two ways around this: either support multiple ssh connections per your "connection" which is messy and complicated, or provide a multiplexing protocol which requires an agent on the target machine, also messy and complicated.
> 5. You will need to support multi-hop connections, where the user connects to an intermediate machine first, and then connects to the target machine from there. Two sets of credentials will need to be provided. Most intermediate systems seem to support tunneling, so you can usually set up a tunnel to the target machine once the connections have been established.

@jarrah42 I would hope to pick these things up from PTP once you’ve finished remodeling it to not use osgi. Let me know if that isn’t going to happen though!

org.eclipse.ice.commands/src/main/java/org/eclipse/ice/commands/RemoteCommand.java
jayjaybillings: @SmithRWORNL How did you do quoted comment of a tiny piece of code? Awesome!

SmithRWORNL: @jayjaybillings Go into Files Changed, hover over the lines you want, and click and drag.

osbornjd:
> * [ ]  You should change the run environment from 1.7 to at least 1.8 but preferrably 11.

Done
> * [ ]  In connection manager, you should examine the case where the connection does not exist in the map. Specifically, your unit test should cover the case where names of nonexistent connections and null are passed into the getConnection function. I see a very trivial test for getConnection() at the moment.

Added a test, the function returns null if the name is not in the hashmap
> * [ ]  Also in connection manager, I believe you can just get the value set for the closeAllConnections function instead of pulling the key set then iterating over get().

Done
> * [ ]  Why did we make ConnectionManager static again? It makes me uneasy.

When we first talked about this design we thought that ConnectionManager should be a static factory, but I can't recall why now. I suppose it doesn't have to be static, so I can change that if we don't have a good reason to leave it as static. 
> * [ ]  What’s the plan to remove ConnectionManager.getPassword() after our discussion yesterday?

I think the plan is to implement the abstract factory class which delegates to the various concrete classes that we discussed. I just haven't had time to think about this more and do it, so I thought I would leave it as is for this PR and then return to it in a future PR since it will be changing a lot of the design of the password approval.
> * [ ]  Why would we expose/share all of our connections with ConnectionManager.getConnectionList()? (I’m genuinely looking for a good reason, and not yet saying remove it.)

My thought was that in the event a developer or user wanted to know what available connections and what their statuses were for e.g. executing new jobs. Since the ssh password isn't kept anywhere (and there are several comments advising not to store it), I thought this would be safe.
> * [ ]  In RemoteCommand, what happened to cancel()?

cancel() is in the Command super class since it applies to both local and remote commands.

> As more of a long-term comment, RemoteCommand should use the RemoteFileHandler to move its files instead of transferring them on its own. But that’s a long term thing. This goes to @jarrah42‘s comment about sftp as well.

Yes, that is on my to do list (I even made a comment about it above the function to remind myself). Need to implement the remote file handling first.

> Also longer term, we need to make sure that RemoteCommandTest runs on Windows since it is using a temp directory. Probably a task for @SmithRWORNL.

Yes, I can talk to Robert about this once the code is in the repo.

Also @jarrah42 I am not ignoring your comments, I think they just require a little more thought on my end. Here are my initial thoughts...

> Here are some things that are going to bite you:
> 
> 1. JSch is not thread safe. So if you're calling any JSch methods from different threads (and it looks like you are) then you'd better be handling the synchronization or you're going to get corruption/weird behavior.

You are right, this is something that Jay and I have discussed and it is on my to do list for making all of the commands thread safe. Initially my thought was that I would fill out the design/functionality of the package, and then implement this.
> 2. JSch is being phased out of Eclipse in favor of Apache Mina as the former is not supported any more.

I'll have to talk to @jayjaybillings about this, we had started using this since it was the most obvious API as it has already been utilized in ICE but if it is being phased out then maybe we should consider moving to Mina.
> 3. Some systems do not allow sftp access, so if you're relying on sftp for file manipulation commands then you should at least test if it is available and notify the user if not. Having an alternative you can use as a fallback is also helpful.

I can work on this, and as @jayjaybillings mentioned it is on my to do list to completely rewrite the file manipulation in RemoteCommand with the RemoteFileHandler once I fill that functionality out.
> 4. Many HPC systems (including OLCF) restrict the number of ssh channels to one, so relying on being able to open channels whenever you want is problematic. There are only two ways around this: either support multiple ssh connections per your "connection" which is messy and complicated, or provide a multiplexing protocol which requires an agent on the target machine, also messy and complicated.

This is very useful to know and will require some more thought on my end...
> 5. You will need to support multi-hop connections, where the user connects to an intermediate machine first, and then connects to the target machine from there. Two sets of credentials will need to be provided. Most intermediate systems seem to support tunneling, so you can usually set up a tunnel to the target machine once the connections have been established.

This is a good suggestion, and I have had something similar/related on my mind. Right now the only tests I have for the remote commands are for interacting via a remote machine with my local machine, but as you say the API should work between two remote machines. This is something else that is on my mind/to-do list.

osbornjd:
@jayjaybillings I believe I have addressed your comments, both about ConnectionManager being local and the input file naming scheme. Take a look at the previous two commits if you have questions.

osbornjd:
### Addition of Remote File Handling

The previous  10 commits or so have completed and filled out the remote file handling source code. The remote file handling API is now functional as indicated in the tests, and can move files locally-to-remotely, remotely-to-locally, and remotely-to-remotely (over the same host). Files moving remotely-to-remotely on some other host is left for future development, as discussed previously in this (now unruly) pull request.

@jayjaybillings I believe this is now ready for you to take a look at as a more complete package for the "remote stuff" API.

jarrah42:
Was there a reason to create your own file manipulation interface? Eclipse already has an IFileStore interface that works very well for remote/local access (https://help.eclipse.org/2019-09/topic/org.eclipse.platform.doc.isv/reference/api/org/eclipse/core/filesystem/package-summary.html?cp=2_1_0_30) . I would suggest using that interface (copying the source if necessary). There are some dependencies on equinox but I would think these could be refactored pretty easily.

osbornjd:
@jarrah42 I can talk to @jayjaybillings about this, I was originally just using the JSch package for the whole remote API part since this is what was the case in previous versions of ICE. I'll look more carefully at the source for the IFileStore interface. You had also mentioned that JSch was being phased out of Eclipse in favor of Apache Mina, so we can discuss this in conjunction with IFileStore.

org.eclipse.ice.commands/src/main/java/org/eclipse/ice/commands/FileHandlerFactory.java
```diff
@@ -45,16 +45,19 @@ public FileHandlerFactory() {
...
+		// Determine if the hostname in the config is local or not
+		boolean isLocal = isLocal(connectionConfig.getHostname());
 
+		// If it is, return a local file handler, otherwise remote
 		if (isLocal) {
 			handler = new LocalFileHandler();
 		} else {
 			handler = new RemoteFileHandler();
+			// Set the connection information for the remote file handler
+			((RemoteFileHandler) handler).setConnectionConfiguration(connectionConfig);
```
jarrah42: Type casting should be avoided and seems indicative of a problem with the model. Why are you not passing the configuration via the constructor?.

org.eclipse.ice.commands/src/main/java/org/eclipse/ice/commands/LocalFileHandler.java
```diff
@@ -32,39 +29,37 @@ public LocalFileHandler() {
...
+	 * See
+	 * {@link org.eclipse.ice.commands.FileHandler#setConfiguration(String, String)}
 	 */
 	@Override
 	protected void configureMoveCommand(final String source, final String destination) {
 		command = new LocalMoveFileCommand();
-		// Cast the Command as a LocalMoveFileCommand to set the source and destination paths
+		// Cast the Command as a LocalMoveFileCommand to set the source and destination
+		// paths
 		((LocalMoveFileCommand) command).setConfiguration(source, destination);
```
jarrah42: Type casting again.

org.eclipse.ice.commands/src/main/java/org/eclipse/ice/commands/LocalFileHandler.java
```diff
@@ -32,39 +29,37 @@ public LocalFileHandler() {
...
+	 * See
+	 * {@link org.eclipse.ice.commands.FileHandler#setConfiguration(String, String)}
 	 */
 	@Override
 	protected void configureCopyCommand(final String source, final String destination) {
 		command = new LocalCopyFileCommand();
-		// Cast the Command as a LocalCopyFileCommand to set the source and destination paths
+		// Cast the Command as a LocalCopyFileCommand to set the source and destination
+		// paths
 		((LocalCopyFileCommand) command).setConfiguration(source, destination);
```
jarrah42: Type casting again.

org.eclipse.ice.commands/src/main/java/org/eclipse/ice/commands/RemoteFileHandler.java
```diff
@@ -21,47 +27,333 @@
...
 	protected void configureMoveCommand(String source, String destination) {
-		// TODO Auto-generated method stub
-		
+		Connection connection = getHandlerConnection();
+		// Now instantiate the command as a RemoteMoveFileCommand
+		command = new RemoteMoveFileCommand();
+
+		// Set the command to have this connection and connection configuration
+		command.setConnectionConfiguration(connection.getConfiguration());
+		((RemoteCommand) command).setConnection(connection);
```
jarrah42: Type casting again.

org.eclipse.ice.commands/src/main/java/org/eclipse/ice/commands/RemoteFileHandler.java
```diff
@@ -21,47 +27,333 @@
...
+
+		// Cast the command as a remote move file command
+		((RemoteMoveFileCommand) command).setConfiguration(source, destination);
```
jarrah42: Type casting again.

org.eclipse.ice.commands/src/main/java/org/eclipse/ice/commands/ConnectionConfiguration.java
```diff
@@ -43,35 +58,104 @@ public ConnectionConfiguration() {
...
+	}
+
+	/**
+	 * A setter to access
+	 * {@link org.eclipse.ice.commands.ConnectionConfiguration#name}
+	 * 
+	 * @return - name
+	 */
+	public void setName(String _name) {
+		name = _name;
```
jarrah42: Inconsistent naming. Use "this.name = name"

org.eclipse.ice.commands/src/main/java/org/eclipse/ice/commands/ConnectionManager.java
```diff
@@ -35,45 +52,235 @@ public ConnectionManager() {
...
+		try {
+			password = in.readLine();
+			in.close();
+		} catch (IOException e) {
+			logger.error("Couldn't read the password...");
+			return null;
+		}
+
+		eraser.stopErasing();
+		System.out.print("\b");
```
jarrah42: Is this required?

org.eclipse.ice.commands/src/main/java/org/eclipse/ice/commands/FileHandler.java
```diff
@@ -192,30 +217,20 @@ protected boolean createDirectories(String dest) throws IOException {
...
 		// Execute the file transfer
 		transferStatus = command.execute();
-
+		
 		// Check that the move succeeded
 		boolean check = false;
 		try {
 			check = exists(destination);
 		} catch (IOException e) {
 			e.printStackTrace();
```
jarrah42: If action is not going to be taken here, then the exception should be passed to the caller.

org.eclipse.ice.commands/src/main/java/org/eclipse/ice/commands/LocalCommand.java
```diff
@@ -115,13 +98,19 @@ protected CommandStatus run() {
...
+			// Finish the job by logging output. If the job is not completed, then
+			// it will be monitored in monitorJob
+			status = finishJob();
 
 			// Check the status to ensure the job hasn't failed
 			try {
 				checkStatus(status);
 			} catch (IOException e) {
+				// If not, return failed
 				e.printStackTrace();
```
jarrah42: Should be logged or not.

org.eclipse.ice.commands/src/main/java/org/eclipse/ice/commands/LocalCommand.java
```diff
@@ -154,6 +145,224 @@ protected CommandStatus run() {
...
+			// Try to get the exit value of the job
+			// If the job completed successfully this will be 0
+			try {
+				exitValue = job.exitValue();
+			} catch (IllegalThreadStateException e) {
+				// Complain, but keep watching
+				try {
+					commandConfig.getStdErr().write(getClass().getName() + "IllegalThreadStateException!: " + e);
+				} catch (IOException e1) {
+					e1.printStackTrace();
```
jarrah42: Should be logged or otherwise handled.

org.eclipse.ice.commands/src/main/java/org/eclipse/ice/commands/LocalCommand.java
```diff
@@ -154,6 +145,224 @@ protected CommandStatus run() {
...
+			try {
+				job.waitFor(1000, TimeUnit.MILLISECONDS);
+				// Try again
+				exitValue = job.exitValue();
+			} catch (InterruptedException e) {
+				// Complain
+				try {
+					commandConfig.getStdErr().write(getClass().getName() + " InterruptedException!: " + e);
+				} catch (IOException e1) {
+					e1.printStackTrace();
```
jarrah42: Should be logged or otherwise handled.

org.eclipse.ice.commands/src/main/java/org/eclipse/ice/commands/LocalCommand.java
```diff
@@ -154,6 +145,224 @@ protected CommandStatus run() {
...
+				logger.info("Job is no longer alive, done monitoring");
+				break;
+			}
+		}
+
+		// Print the final exitValue of the job to the output log file
+		try {
+			commandConfig.getStdOut().write("INFO: Command::monitorJob Message: Exit value = " + exitValue + "\n");
+		} catch (IOException e) {
+			e.printStackTrace();
```
jarrah42: Should be logged or otherwise handled.

org.eclipse.ice.commands/src/main/java/org/eclipse/ice/commands/LocalCommand.java
```diff
@@ -154,6 +145,224 @@ protected CommandStatus run() {
...
+		}
+
+		logger.info("Finished monitoring job with exit value: " + exitValue);
+		if (exitValue == 0)
+			return CommandStatus.SUCCESS;
+		else
+			return CommandStatus.FAILED;
+	}
```
jarrah42: Why is an else clause needed? I think it's clearer to use:

if (exitValue != 0) {
	return CommandStatus.FAILED;
}

return CommandStatus.SUCCESS;

org.eclipse.ice.commands/src/main/java/org/eclipse/ice/commands/LocalCopyFileCommand.java
```diff
@@ -73,11 +72,14 @@ protected CommandStatus run() {
 		// Try to copy from source to destination, overwriting if the file already
 		// exists at the destination. If it can't, complain.
 		try {
+			status = CommandStatus.RUNNING;
 			Files.copy(source, destination.resolve(source.getFileName()), REPLACE_EXISTING);
 		} catch (IOException e) {
+			status = CommandStatus.FAILED;
 			e.printStackTrace();
```
jarrah42: Remove or log.

org.eclipse.ice.commands/src/main/java/org/eclipse/ice/commands/LocalMoveFileCommand.java
```diff
@@ -68,45 +68,62 @@ public CommandStatus execute() {
...
 		// extension
 		// type, then we are just changing a file name
 		if (sameFileExtension) {
 			try {
+				status = CommandStatus.RUNNING;
 				// destinationDirs will have the desired file name in the length-1 entry
 				Files.move(source, source.resolveSibling(destinationDirs[destinationDirs.length - 1]),
 						REPLACE_EXISTING);
 			} catch (IOException e) {
 				e.printStackTrace();
```
jarrah42: Remove or log.

org.eclipse.ice.commands/src/main/java/org/eclipse/ice/commands/LocalMoveFileCommand.java
```diff
@@ -68,45 +68,62 @@ public CommandStatus execute() {
...
 			} catch (IOException e) {
-				e.printStackTrace();
+				try {
+					// If that catches, then try to move it into a different directory
+					// with the same name
+					Files.move(source, destination.resolve(source.getFileName()));
+				} catch (IOException e1) {
+					// If that catches, then it really failed and return as such
+					e.printStackTrace();
+					e1.printStackTrace();
```
jarrah42: Remove these or log the exception.

org.eclipse.ice.commands/src/main/java/org/eclipse/ice/commands/LocalMoveFileCommand.java
```diff
@@ -138,24 +156,14 @@ private boolean checkFileExtension() {
...
-		// If the source and destination have the same path and have the same file extension
-		// then we should treat this as a file name change, otherwise it is a file directory
-		// move
-		if(!different && sameFileExt)
+		// If the source and destination have the same path and have the same file
+		// extension then we should treat this as a file name change, otherwise it is a
+		// file directory move
+		if (!different && sameFileExt)
 			return true;
 		else
```
jarrah42: Don't use else with return.

org.eclipse.ice.commands/src/main/java/org/eclipse/ice/commands/RemoteCommand.java
```diff
@@ -38,63 +69,392 @@ public RemoteCommand() {
...
+				secondConnection = manager.openConnection(extraConnection);
+				// Set the commandConfig hostname to be the extra connection, since this is
+				// really where the job will run
+				commandConfig.setHostname(extraConnection.getHostname());
+			}
+		} catch (JSchException e) {
+			// If the connection(s) can't be opened, we can't be expected to execute a job
+			// remotely!
+			status = CommandStatus.INFOERROR;
+			e.printStackTrace();
```
jarrah42: Remove or log.

org.eclipse.ice.commands/src/main/java/org/eclipse/ice/commands/RemoteCommand.java
```diff
@@ -38,63 +69,392 @@ public RemoteCommand() {
...
+			try {
+				stdErrStream = new FileOutputStream(commandConfig.getErrFileName(), true);
+				stdOutStream = new FileOutputStream(commandConfig.getOutFileName(), true);
+				BufferedOutputStream stdOutBufferedStream = new BufferedOutputStream(stdOutStream);
+				// Give the streams to the channel now that their names are appropriately set
+				connection.getChannel().setOutputStream(stdOutBufferedStream);
+				((ChannelExec) connection.getChannel()).setErrStream(stdErrStream);
+			} catch (FileNotFoundException e) {
+				// If logging streams can't be set, puke
+				e.printStackTrace();
```
jarrah42: The exception needs to be dealt with or passed up.

org.eclipse.ice.commands/src/main/java/org/eclipse/ice/commands/RemoteCopyFileCommand.java
```diff
@@ -23,55 +25,149 @@
...
+		String command = "cp " + source + " " + destination;
+		// Set the command for the JSch connection
+		((ChannelExec) getConnection().getChannel()).setCommand(command);
+		// If the channel isn't connected, connect and run the command
+		if (!getConnection().getChannel().isConnected()) {
+			getConnection().getChannel().connect();
+		} else {
+			logger.error("Channel isn't connected and can't copy remote to remote...");
+			throw new JSchException();
+		}
```
jarrah42: This style is needlessly complicated. It's much clearer to do:

```
if (getConnection().getChannel().isConnected()) {
    logger.error("Channel isn't connected and can't copy remote to remote...");
    throw new JSchException();
}

getConnection().getChannel().connect();
```

org.eclipse.ice.commands/src/main/java/org/eclipse/ice/commands/ConnectionConfiguration.java
osbornjd: Okay, I formatted all the setter functions the same as you suggested.

org.eclipse.ice.commands/src/main/java/org/eclipse/ice/commands/ConnectionManager.java
osbornjd: Yes, it stops the thread from constantly deleting the previous character when someone is inputting their password (so their password doesn't just get typed at the shell, for example). I'll add a comment.

org.eclipse.ice.commands/src/main/java/org/eclipse/ice/commands/FileHandler.java
osbornjd: Moved the exception checking up to the publicly called function.

org.eclipse.ice.commands/src/main/java/org/eclipse/ice/commands/LocalCommand.java
osbornjd: It is logged in checkStatus

jarrah42: you should also remove all references to e.printStackTrace().

jarrah42: also, if the exception is handled in checkStatus then it should not rethrow the exception only to be ignored.

org.eclipse.ice.commands/src/main/java/org/eclipse/ice/commands/LocalCopyFileCommand.java
osbornjd: See above comment, I went through and made sure every exception was logged.

org.eclipse.ice.commands/src/main/java/org/eclipse/ice/commands/LocalMoveFileCommand.java
osbornjd: See above comment, I went through and made sure every exception was logged.

org.eclipse.ice.commands/src/main/java/org/eclipse/ice/commands/LocalMoveFileCommand.java
osbornjd: See above comment, went through and made sure all exceptions were logged.

org.eclipse.ice.commands/src/main/java/org/eclipse/ice/commands/LocalMoveFileCommand.java
osbornjd: See above comment, modified all cases where else was used with return.

org.eclipse.ice.commands/src/main/java/org/eclipse/ice/commands/RemoteCommand.java
osbornjd: See above.

org.eclipse.ice.commands/src/main/java/org/eclipse/ice/commands/RemoteCommand.java
osbornjd: I've looked through all classes to make sure that all exceptions are properly dealt with.

org.eclipse.ice.commands/src/main/java/org/eclipse/ice/commands/RemoteCopyFileCommand.java
osbornjd: I altered these were relevant to be clearer.

org.eclipse.ice.commands/src/main/java/org/eclipse/ice/commands/LocalCommand.java
osbornjd: > you should also remove all references to e.printStackTrace().

Why should I do this? Isn't this useful for debugging purposes?

> also, if the exception is handled in checkStatus then it should not rethrow the exception only to be ignored.

checkStatus logs whether or not the status is failed and then throws an IOException, which is why if checkStatus is caught the run function returns CommandStatus.FAILED. The exception isn't handled in checkStatus, it is just logged, and then handled in the catch to checkStatus by returning that the job has failed.

jarrah42: > > you should also remove all references to e.printStackTrace().
> 
> Why should I do this? Isn't this useful for debugging purposes?
> 

No, because you can't guarantee that stdout is going to go somewhere useful, and because it can't be controlled. Also, since you're ignoring the exception, the user is going to see this when something fails have no idea what is going on. It is better to always use a logging framework for debugging so it is only visible to the developer.

> > also, if the exception is handled in checkStatus then it should not rethrow the exception only to be ignored.
> 
> checkStatus logs whether or not the status is failed and then throws an IOException, which is why if checkStatus is caught the run function returns CommandStatus.FAILED. The exception isn't handled in checkStatus, it is just logged, and then handled in the catch to checkStatus by returning that the job has failed.

I'm not sure why you'd want to throw an exception at this point then, rather than just return a boolean. Particularly since you're just essentially ignoring the exception anyway.

osbornjd: > > > also, if the exception is handled in checkStatus then it should not rethrow the exception only to be ignored.
> > 
> > 
> > checkStatus logs whether or not the status is failed and then throws an IOException, which is why if checkStatus is caught the run function returns CommandStatus.FAILED. The exception isn't handled in checkStatus, it is just logged, and then handled in the catch to checkStatus by returning that the job has failed.
> 
> I'm not sure why you'd want to throw an exception at this point then, rather than just return a boolean. Particularly since you're just essentially ignoring the exception anyway.

Okay, that is fair. I've just changed the function to a return a boolean, and if the status is bad the job is marked as failed.

> > > you should also remove all references to e.printStackTrace().
> > 
> > 
> > Why should I do this? Isn't this useful for debugging purposes?
> 
> No, because you can't guarantee that stdout is going to go somewhere useful, and because it can't be controlled. Also, since you're ignoring the exception, the user is going to see this when something fails have no idea what is going on. It is better to always use a logging framework for debugging so it is only visible to the developer.

Okay, I have gone through and removed all instances where an exception is thrown to print the stack trace.

> > > also, if the exception is handled in checkStatus then it should not rethrow the exception only to be ignored.
> > 
> > 
> > checkStatus logs whether or not the status is failed and then throws an IOException, which is why if checkStatus is caught the run function returns CommandStatus.FAILED. The exception isn't handled in checkStatus, it is just logged, and then handled in the catch to checkStatus by returning that the job has failed.
> 
> I'm not sure why you'd want to throw an exception at this point then, rather than just return a boolean. Particularly since you're just essentially ignoring the exception anyway.

osbornjd:
I believe I have addressed the comments left by @jarrah42, other than the type cast one which is waiting on @jayjaybillings to comment on.

org.eclipse.ice.commands/src/main/java/org/eclipse/ice/commands/FileHandlerFactory.java
jayjaybillings: In this particular case, just do this:
```java
RemoteFileHandler remoteHandler = new RemoteFileHandler();
remoteHandler.setConnectionConfiguration(connectionConfig);
handler = remoteHandler;
```

jayjaybillings:
> Was there a reason to create your own file manipulation interface? Eclipse already has an IFileStore interface that works very well for remote/local access (https://help.eclipse.org/2019-09/topic/org.eclipse.platform.doc.isv/reference/api/org/eclipse/core/filesystem/package-summary.html?cp=2_1_0_30) . I would suggest using that interface (copying the source if necessary). There are some dependencies on equinox but I would think these could be refactored pretty easily.

We're specifically avoiding OSGi code. It has too much baggage. I'm not opposed to taking inspiration from it though.

jayjaybillings:
I just want to say a quick thanks to @jarrah42 and @SmithRWORNL for their excellent review of this PR. Great work guys! This is probably our best reviewed PR ever (poor @osbornjd! ;)).

org.eclipse.ice.commands/src/main/java/org/eclipse/ice/commands/FileHandlerFactory.java
osbornjd: Okay I see, I will change this and the other type casting instances to something similar

jarrah42:
> > Was there a reason to create your own file manipulation interface? Eclipse already has an IFileStore interface that works very well for remote/local access (https://help.eclipse.org/2019-09/topic/org.eclipse.platform.doc.isv/reference/api/org/eclipse/core/filesystem/package-summary.html?cp=2_1_0_30) . I would suggest using that interface (copying the source if necessary). There are some dependencies on equinox but I would think these could be refactored pretty easily.
> 
> @jarrah42 We're specifically avoiding OSGi code. It has too much baggage. I'm not opposed to taking inspiration from it though.

It might be worth a discussion if we're interested in pursuing the micro services approach: a) there are some really interesting frameworks coming out that avoid a lot of the Java EE baggage (e.g. https://microprofile.io); b) there is a new effort called OSGi Connect aimed at dealing better with non-OSGi/OSGi issues; c) JPMS is really screwed up (e.g. no runtime components); and d) the OSGi frameworks are pretty lean now. Anyway, just my 2c worth.

osbornjd:
> * [ ]  It looks like the bin/ directory of org.eclipse.ice.commands.swt is not properly ignored.

I'll have to ask you more about this, because I'm not sure what you mean.

> * [ ]  The readme should note that mvn clean install will install to the local repository in ~/.m2, and that it is also possible to build without installing by running mvn clean verify.

Added to the README.

> * [ ]  I think we should figure out how to fix the log4j dependency issues instead of working around it.

Okay, I'll have to talk to Lance about this since he was originally the one who helped solve the issue in the first place with Jean.

> * [ ]  Should build.sh do clean verify or clean install? You probably want to rename it to "build-verify" or "build-install" depending on your intent.

I just removed this, since it is invalid now. Originally this was for Caleb, but it is obsolete now that the README was added with build instructions.

> * [ ]  In CommandConfiguration.java, shouldn't we also write the stack trace to the log?

@jarrah42 suggested I remove all printStackTrace from any exception/error catching, so I removed all of these. I think it makes for a cleaner output and each exception is logged with a message for debugging purposes.

> * [ ]  Should Connection.java be separate into an IConnection interface and a JSchConnection implementation?

I guess this is forward thinking in the event that someone would want to implement some other kind of connection if e.g. a different connection API from JSch wanted to be used? At the moment it doesn't seem necessary to me since this class is kind of boring/sparse already, but maybe this is what you were thinking about when making this suggestion.

> * [ ]  In ConnectionConfiguration.java, get/setDeleteWorkingDirectory() can just be deleteWorkingDirectory(). That is a cleaner way to handle boolean functions (no need for get/set).

Modified as suggested.

> * [ ]  You can do the same trick to avoid the cast in LocalFileHandler that I suggested for the factory in @jarrah42's question.

You are right, I changed all of these. Not sure why I didn't see that as a possibility before...

> * [ ]  I believe the same trick works in the configureCopy/MoveCommand functions in RemoteFileHandler.

Ditto to above.

> * [ ]  I saw this code in several places:
> 
> ```java
> 			// Determine the move type and, thus, how to move the file
> 			if (moveType == 1) { // If move type is local -> remote, use put
> 				channel.put(source, destination);
> 			} else if (moveType == 2) { // if move type is remote -> local, use get
> 				channel.get(source, destination);
> 			} else if (moveType == 3) { // if move type is remote -> remote, call function
> 				moveRemoteToRemote();
> 			} else {
> 				logger.error("Unknown move type...");
> 				status = CommandStatus.FAILED;
> 				return status;
> 			}
> 			// If permissions was actually instantiated and isn't the default, then perform
> 			// a chmod
> 			if (permissions != -999)
> 				channel.chmod(permissions, destination);
> ```
> 
> We should move it its own class if we are going to reproduce it and require it in more than one place.

Moved this to its own class, the implementation is cleaner now.

> * [ ]  Is the VaadinConnectionAuthorizationHandler an unused stub? If so, should we remove it? Same with it's test.

At the moment, yes. I put these there when designing the authorization structure since this was something that we talked about it. I have removed it for now since I need to work with Lance on implementing this in the longer term.

> * [ ]  You may want to consider adding a method to CommandFactoryTest that creates a default config for you that you then modify for your needs in the test. It would in principle be shorter for some cases than reproducing the whole big block of construction code.

Yes, I thought about this, but I also thought it would be nice if the tests (especially the top most factory tests, which are most relevant to users) would act as standalone examples. I refactored the code around so that this is still the case but that the tests reduce some of the repetition in the command configuration block.

> * [ ]  One other thing that wasn't clear, or at least I thought I saw it and then lost track of it: How can clients register their own ConnectionAuthorizationHandlers with the factory? Is this available?

I made a first pass at this functionality and added it. I'll have to think about some of the finer details further in the future, but for now this functionality is available.

> Overall, I think this is looking very, very good. Thanks @osbornjd!

Thanks!

jayjaybillings:
To address the comments from @osbornjd above:
1) Let me know what Lance says.
2) Yes, that was more of a forward thinking comment. If it was a little bit of work, that would be OK to do now, but I understand if you just want to make a separate ticket.

With respect to the exceptions, is the full text of each exception being logged?

osbornjd:
> To address the comments from @osbornjd above:
> 
> 1. Let me know what Lance says.

Will do.

> 2. Yes, that was more of a forward thinking comment. If it was a little bit of work, that would be OK to do now, but I understand if you just want to make a separate ticket.

I'm just going ahead and working on it now. After thinking about it for a bit I thought it would be straightforward and I think it is, there are just a few other things to work out.

> With respect to the exceptions, is the full text of each exception being logged?

The text of the exception isn't necessarily being logged, but there is text logged for each exception. I can go through and also include the actual kind of exception that it is in the logs (e.g. IOException, SftpException, whatever).

osbornjd:
@jayjaybillings Lance and I discussed that this issue is not present, and shouldn't be when the code is put into the mvnrepository. I updated the README to reflect this, and pending other comments I think this PR should be in good shape.

org.eclipse.ice.commands/src/main/java/org/eclipse/ice/commands/RemoteTransferExecution.java
```diff
@@ -76,7 +76,7 @@ protected CommandStatus executeTransfer(Connection connection, String source, St
 				channel.chmod(permissions, destination);
```
jayjaybillings: This is a situation where I would like to see something like
```java
logger.error(e.printStackTrace());
```
or
```java
logger.catching(e);
```

osbornjd:
> Is all the Windows code under unit test?

Not yet - probably better to just wait for me to finish this. I have all the tests passing on windows also except for the remote command stuff.

> I also included an example with the exception thing.
I'll add some stack trace logging in addition to finishing the windows tests

osbornjd:
For documentation purposes, this PR should now address issue [412](https://github.com/eclipse/ice/issues/412) which requested windows testing capabilities for the commands API.

Issue #413, opened 2019-11-06
Commands File Browsing


osbornjd:
Commands (in particular FileHandler) could benefit from a feature where file browsing the local/remote system is possible. Files and directories should be accessible through this API.

PR #414, opened 2019-11-08
File Browser for commands API

osbornjd:
### Summary
This PR introduces a file browsing API within the Commands API, in reference to [issue 413](https://github.com/eclipse/ice/issues/413). The `FileBrowser` is designed to work in hand with the `IFileHandler` class. Tests are introduced that build a directory/file structure underneath a new directory, and then look for and confirm that the `FileBrowser` returns the files/directories that are created (depending on the test).

### Design 
A `FileBrowser` interface is introduced, from which a local and remote `FileBrowser` implement from. These subclasses contain the logic for the file browsing API on the local or remote host. They are called from the FileHandler like the other abstract `IFileHandler` classes, for example, supposing you have an object `handler` of type `FileHandler`
```java
ArrayList<String> files = handler.listFiles("/some/top/directory");
```
The function returns a vector of strings containing the paths of all of the files below the given directory. Similarly, one could ask for the directories by calling
```java
ArrayList<String> directories = handler.listDirectories("/some/top/directory");
```
to be given the directories underneath the given top directory.


org.eclipse.ice.commands/src/main/java/org/eclipse/ice/commands/FileBrowser.java
```diff
@@ -0,0 +1,58 @@
...
+ * 
+ * @author Joe Osborn
+ *
+ */
+public interface FileBrowser {
+
+	/**
+	 * Logger for handling event messages and other information.
+	 */
+	static final Logger logger = LoggerFactory.getLogger(LocalFileBrowser.class);
```
jayjaybillings: LocalFileBrowser.class?

jayjaybillings:
I made some comments, but @osbornjd you're a committer now so I leave it to you to make the changes and merge or just add tickets and come back later. Good work.

org.eclipse.ice.commands/src/main/java/org/eclipse/ice/commands/FileBrowser.java
osbornjd: Whoops... good catch. Fixed

osbornjd:
> This all looks fine to me, but here are some design points to consider:
> 
> 1. Why not have a method call IFileHandler.getFileBrowser() that returns a FileBrowser? This would eliminate the delegation code on FileHandlers and just let clients work directly with the FileBrowsers.

This can be done, but I don't understand the benefit of interacting with the FileBrowser directly since the FileHandler is still the top most class that clients interact with. Why wouldn't we have FileHandler delegate to the FileBrowser so that the FileHandler takes care of all of your file-ing needs, e.g. transferring or browsing? If the FileHandler returned a FileBrowser, that is another line of code for the client, e.g.

```java
FileHandler handler = factory.getFileHandler();
FileBrowser browser = handler.getBrowser(); // this line would be extra, right?
browser.doStuff();
```

Can you comment on why this suggestion would be better?

> 2. I like the separation between directories and files on the FileBrowser classes. This basically gives us a file system tree. We should consider whether or not this is optimal though and if a real, lazy tree is needed. One thing that would be interesting is if someone wanted to do a recursive scan of a directory. How would they do it?

That is a good suggestion. I think you are right that a default version should be implemented so that one can get everything underneath a particular top directory without differentiating between files and directories, if I understand your comment correctly.

jayjaybillings:
> > This all looks fine to me, but here are some design points to consider:
> > 
> > 1. Why not have a method call IFileHandler.getFileBrowser() that returns a FileBrowser? This would eliminate the delegation code on FileHandlers and just let clients work directly with the FileBrowsers.
> 
> This can be done, but I don't understand the benefit of interacting with the FileBrowser directly since the FileHandler is still the top most class that clients interact with. Why wouldn't we have FileHandler delegate to the FileBrowser so that the FileHandler takes care of all of your file-ing needs, e.g. transferring or browsing? If the FileHandler returned a FileBrowser, that is another line of code for the client, e.g.
> 
> ```java
> FileHandler handler = factory.getFileHandler();
> FileBrowser browser = handler.getBrowser(); // this line would be extra, right?
> browser.doStuff();
> ```
> 
> Can you comment on why this suggestion would be better?

Encapsulation. What if I want to give some client code the ability to browse, but not to move files? Think about the differences in activities that this class can do:
```java
public MyClass(IFileHandler handler)
```
vs
```java
public MyClass(IFileBrowser browser)
```
There are other design points, but I would put this one up for debate first.

> > 1. I like the separation between directories and files on the FileBrowser classes. This basically gives us a file system tree. We should consider whether or not this is optimal though and if a real, lazy tree is needed. One thing that would be interesting is if someone wanted to do a recursive scan of a directory. How would they do it?
> 
> That is a good suggestion. I think you are right that a default version should be implemented so that one can get everything underneath a particular top directory without differentiating between files and directories, if I understand your comment correctly.

That wasn't what I was getting at. What I'm saying is that it is good that we have them separated. However, I'm wondering
1) If the current implementation has good performance in memory and access times?
2) If people can do more complicated things like recursive scans?

Here's a case study: If I have a directory with over 300k files (and I have a lot of those kinds of directories...), and some random distribution among subdirectories, can I recursively walk the whole tree efficiently?

This is something to consider, but perhaps not something that we immediately implement.

osbornjd:
> > > This all looks fine to me, but here are some design points to consider:
> > > 
> > > 1. Why not have a method call IFileHandler.getFileBrowser() that returns a FileBrowser? This would eliminate the delegation code on FileHandlers and just let clients work directly with the FileBrowsers.
> > 
> > 
> > This can be done, but I don't understand the benefit of interacting with the FileBrowser directly since the FileHandler is still the top most class that clients interact with. Why wouldn't we have FileHandler delegate to the FileBrowser so that the FileHandler takes care of all of your file-ing needs, e.g. transferring or browsing? If the FileHandler returned a FileBrowser, that is another line of code for the client, e.g.
> > ```java
> > FileHandler handler = factory.getFileHandler();
> > FileBrowser browser = handler.getBrowser(); // this line would be extra, right?
> > browser.doStuff();
> > ```
> > 
> > 
> > Can you comment on why this suggestion would be better?
> 
> Encapsulation. What if I want to give some client code the ability to browse, but not to move files? Think about the differences in activities that this class can do:
> 
> ```java
> public MyClass(IFileHandler handler)
> ```
> 
> vs
> 
> ```java
> public MyClass(IFileBrowser browser)
> ```
> 
> There are other design points, but I would put this one up for debate first.

Okay, I believe my last commit has implemented this. I also adjusted the tests to be a little cleaner/separated into FileBrowser tests. 


> > > 1. I like the separation between directories and files on the FileBrowser classes. This basically gives us a file system tree. We should consider whether or not this is optimal though and if a real, lazy tree is needed. One thing that would be interesting is if someone wanted to do a recursive scan of a directory. How would they do it?
> > 
> > 
> > That is a good suggestion. I think you are right that a default version should be implemented so that one can get everything underneath a particular top directory without differentiating between files and directories, if I understand your comment correctly.
> 
> That wasn't what I was getting at. What I'm saying is that it is good that we have them separated. However, I'm wondering
> 
> 1. If the current implementation has good performance in memory and access times?
> 2. If people can do more complicated things like recursive scans?
> 
> Here's a case study: If I have a directory with over 300k files (and I have a lot of those kinds of directories...), and some random distribution among subdirectories, can I recursively walk the whole tree efficiently?
> 
> This is something to consider, but perhaps not something that we immediately implement.

I see what you mean. Locally I think this would be fine, but remotely it isn't clear to me and I'd need to build a test case to see how inefficient it is. I would agree that we push this to a ticket and merge this as is for the time being.

@jayjaybillings I will go ahead and merge this as it is now and if you have comments/suggestions on the FileBrowser encapsulation I can implement that in a new ticket.

PR #417, opened 2019-12-12
Jump connection commands

osbornjd:
## Summary
This PR introduces the possibility to execute commands over a jump connection, where a user connects to a remote host and then runs a job from that remote host to an alternative remote host. This idea was first introduced in a comment from an earlier [PR 411](https://github.com/eclipse/ice/pull/411) Some design changes were implemented to the code so that RemoteCommand could process both single commands and jump connection commands. In the following discussion, let's assume a situation as follows:

System A (local) ---> System B (jump connection) ---> System C (final remote destination)

Here, the job is being executed from System A, the job files (e.g. scripts, input files, etc) exist on System B, and the job should be run on System C.

The major changes implemented are in `transferFiles()` in `RemoteCommand` and a new function in `ConnectionManager` which allows `Connections` to be opened via a forwarded port. Tests are implemented and need to be automated once an additional dummy server is created. 

Note that many files were changed as I did a sweep of the code and added comments where they were lacking.

### Overview of Design

The jump connection commands works with the following design.

#### Forwarded Connections
A forwarding connection is established with the jump host that operates from the local host through the jump host to the final remote host. This is performed since it is assumed that login credentials to the final host are not available on the local host (otherwise the job could just be run as a normal command). The forwarding operates through a random port on the jump connection, assigned while connecting, to an assumed port 22 on the final remote host. This connection is then used to execute the command.

#### Command Logic

`JSch` does not allow commands to be executed from the jump host to the final remote host (e.g. from System B to System C in the example). Therefore, the jump commands implementation downloads the necessary files from the jump host to the local host (System B to System A). Then, these files are transferred through the forwarded connection from System A directly to System C. This simplifies the implementation in `RemoteCommand` anyway since the only function that needs to be modified is `transferFiles`, after which the execution of the Command can be treated as a normal `RemoteCommand` between System A and System C utilizing the forwarding connection. The files are copied from the jump host to the local host in a temporary folder on the local host, which is deleted after the files are copied across the forwarded connection.


osbornjd:
Closing PR because of mina-sshd updates

PR #418, opened 2020-01-22
Updating jay/MarkIII

osbornjd:
Updating jay/MarkIII with some bug fixes that have been in the osbornjd/ice fork for a while now


osbornjd:
Merging this since a few bug fixes were implemented a while ago and have not been merged into the main repository

PR #419, opened 2020-01-24
Committing hello world modeling project

osbornjd:
Creating a bare bones hello world maven project to work out of
Signed-off-by: Joe Osborn <osbornjd@ornl.gov>


osbornjd:
Merging without review - just a hello world maven project so nothing of interest in here at the moment

PR #420, opened 2020-02-17
Jay/mark iii

hetrickjm:
Adding initial UML model and generated java code from my fork.


org.eclipse.ice.modeling/src/main/java/org/eclipse/ice/modeling/IWorkflow.java
```diff
@@ -0,0 +1,13 @@
+package org.eclipse.ice.modeling;
+
+import org.eclipse.ice.modeling.workflowEngine.*;
+
+public interface IWorkflow {
+
+	/**
+	 *
```
osbornjd: In general within ICE we say that comments for functions should give some brief description of what the function does

org.eclipse.ice.modeling/src/main/java/org/eclipse/ice/modeling/TestDriver_WF.java
```diff
@@ -0,0 +1,26 @@
+package org.eclipse.ice.modeling;
+
+import org.eclipse.ice.modeling.workflowEngine.*;
+import org.eclipse.ice.modeling.actors.*;
+
+public class TestDriver_WF {
+	/**
+	 * @param args
+	 */
+	public static void main(String[] args) {
```
osbornjd: Why is this public static void main()? Why will this need to be run on its own?

org.eclipse.ice.modeling/src/main/java/org/eclipse/ice/modeling/states/State.java
```diff
@@ -0,0 +1,10 @@
+package org.eclipse.ice.modeling.states;
+
+/**
```
osbornjd: We should coordinate this with the [Commands#CommandStatus Enum](https://github.com/eclipse/ice/blob/jay%2FMarkIII/org.eclipse.ice.commands/src/main/java/org/eclipse/ice/commands/CommandStatus.java), so that we have a consistent set of states that every step of the workflow can take. We can modify Commands and this so that they are consistent with one another.

org.eclipse.ice.modeling/src/main/java/org/eclipse/ice/modeling/workflowEngine/Action.java
```diff
@@ -0,0 +1,18 @@
+package org.eclipse.ice.modeling.workflowEngine;
+
+public class Action {
```
osbornjd: As we discussed, we should work to interface this with the Commands package since Commands will do most of what we need Action to do

org.eclipse.ice.modeling/src/main/java/org/eclipse/ice/modeling/workflowEngine/Message.java
```diff
@@ -0,0 +1,78 @@
+package org.eclipse.ice.modeling.workflowEngine;
+
+public class Message {
+
+	/**
+	 * DEPRECATE - This is part of the DataSet MetaData
```
osbornjd: Should this be here, if it is a part of MetaData? Or is this something else that is separate from MetaData?

org.eclipse.ice.modeling/src/main/java/org/eclipse/ice/modeling/workflowEngine/Procedure.java
```diff
@@ -0,0 +1,88 @@
+package org.eclipse.ice.modeling.workflowEngine;
+
+/**
+ * The Procedure hold the set of steps that constitute the workflow.  
+ * My current thinking is that this class does not hold state, only Steps
+ * and Completion Criteria
+ */
+public class Procedure {
+
+	private Step[] steps;    // This should really be a List where steps can be added dynamically
```
osbornjd: I agree that this should be a list, we may also want to consider making it a hashmap if there are reasons to be able to identify individual steps quickly via a string identifier or something like this.

org.eclipse.ice.modeling/src/main/java/org/eclipse/ice/modeling/workflowEngine/Procedure.java
```diff
@@ -0,0 +1,88 @@
...
+
+	public Procedure() {
+		System.out.println("Procedure() constructor");
+		
+		// Create a default set of steps
+		// This should really be a List where steps can be added dynamically
+		numSteps = 3;
+		steps = new Step[numSteps];
+		System.out.println( "##### BEGIN:  CREATE SETPS FOR PROCEDURE " );
+		for (int i=0; i < 3; i++) {
```
osbornjd: I think we shouldn't have this in the constructor at all, and just expect steps to be added dynamically. That way there is no confusion about what a procedure actually contains when you instantiate one, and it is a requirement that you manually add steps e.g.
```
Procedure procedure = new Procedure();
procedure.addStep(step1);
procedure.addStep(step2);
...
```

org.eclipse.ice.modeling/src/main/java/org/eclipse/ice/modeling/workflowEngine/Step.java
```diff
@@ -0,0 +1,56 @@
+package org.eclipse.ice.modeling.workflowEngine;
+
+public class Step {
+
+	private Action action;
```
osbornjd: If a step has one action, what is differentiating it from an action other than the SuccessCriteria? Is Step just a stateful implementation of Action? I need to think through this also.

org.eclipse.ice.modeling/src/main/java/org/eclipse/ice/modeling/workflowEngine/SuccessCriteria.java
```diff
@@ -0,0 +1,37 @@
+package org.eclipse.ice.modeling.workflowEngine;
+
+public class SuccessCriteria {
+
+	private String successMsg;
```
osbornjd: How is this class different from State? Are they doing the same thing, and if not what is their difference?

org.eclipse.ice.modeling/src/main/java/org/eclipse/ice/modeling/workflowEngine/Workflow.java
```diff
@@ -0,0 +1,95 @@
+package org.eclipse.ice.modeling.workflowEngine;
+
+/**
+ * Workflow is the entity that is responsible for managing the processing of a data set.
+ * It is an abstract class which will be instantiated by a concrete class.  It is part
+ * of composition pattern so clients only work with the Workflow interface and do not
+ * need to know anything about the concrete classes or their interfaces.
+ */
+public class Workflow {
```
osbornjd: Add abstract here if Workflow is an abstract class (which I agree it should be)

jayjaybillings:
So, what's this for?

org.eclipse.ice.modeling/src/main/java/org/eclipse/ice/modeling/TestDriver_WF.java
hetrickjm: Just for my testing.  This should disappear once I learn more about how this is going to run and how to make that happen.

org.eclipse.ice.modeling/src/main/java/org/eclipse/ice/modeling/states/State.java
hetrickjm: Yes, and I need to learn how to do enums in Java.  This was just an initial place holder so I did not loose my thinking.

org.eclipse.ice.modeling/src/main/java/org/eclipse/ice/modeling/workflowEngine/Action.java
hetrickjm: Agree

org.eclipse.ice.modeling/src/main/java/org/eclipse/ice/modeling/workflowEngine/Message.java
hetrickjm: There is actually a large discussion that is needed around messages, what they contain, etc.  Specifically, as I have not real had information yet, this is a place where my thinking is evolving from just having some attributes in a message to understanding that the same information may be contained in the meta data.  So there are two thing going on here.  1) this is a concept in the process of changing and 2) we need to start getting specific about the data model and what data we are actually working with.  I only have notional concepts that tend to change every meeting.

org.eclipse.ice.modeling/src/main/java/org/eclipse/ice/modeling/workflowEngine/Procedure.java
hetrickjm: Agree, here again my lack of familiarity and comfort with Java is limiting.  I did this just to prove concepts.  You will see this in multiple places.  I too was thinking hash map, but from what little I read in Java, List is an interface that works with hash maps and other types of Collections, etc.  Just don't know how to work with them yet.

org.eclipse.ice.modeling/src/main/java/org/eclipse/ice/modeling/workflowEngine/Procedure.java
hetrickjm: Totally agree, most of this is just to get something running enough to show myself that a chain of events is possible and instantiating classes is possible.  Note there are only print linesa dn no real logic in anything.  All of this needs to mature and be done correctly.  1) this is still exploratory and 2) I don't quite have all those skills yet, and 3) still have loads of questions around how to structure and do things

org.eclipse.ice.modeling/src/main/java/org/eclipse/ice/modeling/workflowEngine/Step.java
hetrickjm: Good question and not sure yet.  Need to have a better idea of what the processing is for this stuff and run some scenarios.  In the abstract though we need to keep and use the same structure if a step has one action or several.  I actually see an action as being fairly atomic, at least from a workflow standpoint.  Otherwise it kind of gets into being another layer of workflow.  I am totally unclear at this point on much of this and need to work through with you.  I am not clear what all state even needs to be captured at the Step (Task) level.  My gut tells me there is and that is supported by Jay's ontology, but I am way lacking in real detail and understanding at this point.

org.eclipse.ice.modeling/src/main/java/org/eclipse/ice/modeling/workflowEngine/SuccessCriteria.java
hetrickjm: Not sure yet.  Once again, there are a number of things running around in my head that have not crystalized yet.  Initially, in the context of AR, this was to hold a set of "Success view" messages that Mat showed us in one of the message tables he presented.  So I am not yet solid if it is that or some larger context.  It may be possible to handle this with Step State.  Another confusion for me is if success and completion are the same thing.  I am not certain they are.

org.eclipse.ice.modeling/src/main/java/org/eclipse/ice/modeling/workflowEngine/Workflow.java
hetrickjm: Agree.  I think there are some abstract methods along the way too.  However, my familiarity with using "abstract" particularly with Java and somewhat in general is not real strong.  (Knowledge that has long since fled my brain.)

org.eclipse.ice.modeling/src/main/java/org/eclipse/ice/modeling/IWorkflow.java
jayjaybillings: Agreed. Classes and interfaces should also have documentation.

org.eclipse.ice.modeling/src/main/java/org/eclipse/ice/modeling/TestDriver_WF.java
jayjaybillings: Can you make this an actual junit test?

org.eclipse.ice.modeling/src/main/java/org/eclipse/ice/modeling/states/State.java
jayjaybillings: @osbornjd We should ask ourselves whether or not it is necessary to reuse the CommandStatus enum from Commands for two reasons:
1) That's not a lot of code to justify a whole set of dependencies.
2) It may be that they enumerations are close but not exact.

org.eclipse.ice.modeling/src/main/java/org/eclipse/ice/modeling/workflowEngine/Action.java
jayjaybillings: Definitely agree on this part, but the translation is not 1-1. For example, what if it is a native function execution? Then is it really a call to Commands? (I don't know the answer - just saying we should think about it.)

osbornjd:
@jayjaybillings This was just an initial PR to get some structure of code from the UML model we discussed last week with Mat into the repo. @hetrickjm and I discussed a bit offline and we are happy to either merge after the cleanups discussed here or postpone until later. The point of the PR was just to get an outline of code in place based on our discussions last week and get John's ICE development workflow (ha) setup and functional.

jayjaybillings:
I'll catch up with @osbornjd and @hetrickjm about this PR.

jayjaybillings:
What can we do about this PR?

I would suggest that we close it since the code base has moved beyond it (see Tasks, DataElements), but that we keep a copy around in case we want to look at something. Any thoughts?

hetrickjm:
> What can we do about this PR?
> 
> I would suggest that we close it since the code base has moved beyond it (see Tasks, DataElements), but that we keep a copy around in case we want to look at something. Any thoughts?

I agree.  This will not go anywhere at least in this form.  We will come back to autoreduction redesign at a future time as well as workflow.  My guess is this is more likely to emanate from ICE3 in the future

osbornjd:
I also agree on closing it, but with the intent for archiving. If we want to come back and look at the model that we were originally thinking, this could be useful to come back and look at.

jayjaybillings:
Okay. Since it is already in a branch in @hetrickjm's repo and stored here as a PR, I think it is sufficiently archived. Closing this now. Thanks for your input!

PR #421, opened 2020-02-21
Mina Dependency

osbornjd:
This PR changes Commands from using JSch as the dependency responsible for remote connections to Mina sshd. The PR introduces the changes on a feature branch since the associated tests for several of these features fail by default since there are not two remote dummy hosts setup to accommodate the tests. One can run the tests if they please implementing their own remote hosts that they have access to.

In addition, it also introduces:

1. The ability to forward connections from one remote host through the local host to another remote host and running associated commands in this way, e.g.

- Local host A uses remote host B as a jump host to remote host C. The connection is created from host B through host A to host C, assuming that host A does not have direct access to host C.

2. An example showing how to run a command on a remote host B that subsequently runs a job on remote host C. This could be useful for e.g. transferring files from host B to host C directly (rather than through a forwarded connection as in the above description), or running a job from host B to host C etc.


osbornjd:
Merging into feature branch after windows testing passes

PR #423, opened 2020-06-22
Revamp DataElement Annotations: Redefine entry point

dbluhm:
Instead of:
```java
@DataElement
@DataField(fieldName = "age", fieldType = int.class, docString = "The person's age.")
@DataField(fieldName = "firstName", fieldType = String.class, docString = "The person's first name.")
@DataField(fieldName = "lastName", fieldType = String.class, docString = "The person's last name.")
public interface Person {
}
```

DataElements are created as follows:
```java
@DataElement(name = "Person")
public class PersonSpec {
	/**
	 * The person's age.
	 */
	@DataField.Default("42")
	@DataField private int age;

	/**
	 * The person's first name.
	 */
	@DataField.Default(value = "Bob", isString = true)
	@DataField private String firstName;

	/**
	 * The person's last name.
	 */
	@DataField private String lastName;
}
```

This comes with the following benefits:
- We can generate an interface that correctly extends IDataElement.
- The generated interface also includes getters and setters for its DataFields.
- Modifiers (private, public, static, final, etc.) are preserved.
- Annotations other than DataField and DateField.Default are preserved.
- Doc comments are preserved.

#### Default values for fields
Unfortunately, the default values of a VariableElement (like `age`, `firstName`, and `lastName` in the example above) cannot be retrieved during annotation processing unless they are compiled constants (int, boolean, String, etc.) and the field is set as final.

To accommodate setting a default value for non-final fields, the `@DataField.Default` annotation was introduced which takes a String which is placed verbatim in the generated implementation class as the default value for that field. As a result of this, default values of type String must include quotation marks in the String itself along with double escaping escaped values. To assist in correctly formatting strings, `isString` is provided as a flag for `@DataField.Default` that will cause the string to be passed through `Elements.getConstantExpression()` and be properly escaped for outputting to source code.

For final fields with primitive values, `@DataField.Default` is not needed and the value can be set as normal. If the value is not primitive, `DataField.Default` can be used as described above.

#### TODO: Lombok Annotations
Due to the way Lombok manipulates the AST, some lombok annotations modify the field's name to include the annotation itself. For instance, `@NonNull` causes a VariableElement's name information to include `@lombok.NonNull` in addition to `@NonNull` being present in its annotation list. This causes problems with the way we generate fields as an "annotation" gets placed after the type information which is syntactically incorrect.


jayjaybillings:
So what's your proof that it works?

dbluhm:
Here's a "Person" DataElement (including an Interface, Implementation, and PersistenceHandler) generated from `PersonSpec.java` in `org.eclipse.ice.renderer` (hopefully this is what you were looking for):

https://gist.github.com/dbluhm/fdb6454319a6a8ac911e238015a99388

jayjaybillings:
Yes, that's what I'm looking for. Thanks!

I noticed that PersonPersistenceHandler didn't generate class level documentation. Is that intentional or is it possible to add some basic, even boilerplate documentation there?

dbluhm:
> Yes, that's what I'm looking for. Thanks!
> 
> I noticed that PersonPersistenceHandler didn't generate class level documentation. Is that intentional or is it possible to add some basic, even boilerplate documentation there?

Yeah, we should be able to get some basic class level documentation in there. I'll add that.

dbluhm:
```java
...
import com.mongodb.client.MongoDatabase;

/**
 * This is an implementation of IPersistenceHandler<Person>
 * that satisfies the dependencies of the {@code @Persisted} Annotation and was
 * auto-generated by the ICE Framework.
 */
public class PersonPersistenceHandler implements IPersistenceHandler<Person> {

	/**
	 * The name of the collection backing this PersistenceHandler.
...
```

jayjaybillings:
That works for me. Thanks for the contribution! Accepted.

Per the Eclipse IP Policy, as the submitting committer I can accept this code and commit it to the repository for @dbluhm since we are both employees of the same member company, UT-Battelle LLC. (Case 2 on the Eclipse Legal Process poster.)

PR #424, opened 2020-06-24
Unit tests for DataElementProcessor

dbluhm:
This PR adds unit tests for the `DataElementProcessor` using the compile-testing library.

The compile-testing library works by invoking the Java compiler through `javax.tools`. It creates an in-memory file manager so generated and compiled files aren't written to disk. The facilities it provides to then make assertions about the generated files are done through parsing the generated files and comparing the ASTs. As a result, we can test that annotations cause failures when we expect it to and that it generates files matching (with a high level of granularity) what we would expect it to generate **BUT**, as it does not compile the final generated classes, we cannot test the logic of generated classes in this manner.

Most of the tests use the "golden file" test strategy with the minor modification that the golden files are actually patterns. ASTs are parsed from the golden files/patterns and compared with the ASTs of the files generated from the input. In other words, the pattern AST must be a subgraph of the generated AST. A nuance of this strategy is that if methods are included in the pattern, the entire body of the method in the pattern must match exactly (AST-wise) the entire body of the method in the generated AST. Knowing that this would result in potentially brittle test cases, the patterns do not verify presence of the methods directly in the generated Implementation AST. They are, however, verified in the generated Interface and any methods missing in the implementation that are present in the interface are caught during testing. So by verifying both the methods in the interface and the fields in the implementation, the presence of all fields and methods on the generated implementation is verified.

### DIFF
--- a/org.eclipse.ice.dev.annotations/pom.xml
+++ b/org.eclipse.ice.dev.annotations/pom.xml
@@ -88,6 +88,12 @@
 			<scope>test</scope>
 		</dependency>
 		<dependency>
+			<groupId>com.google.testing.compile</groupId>
+			<artifactId>compile-testing</artifactId>
+			<version>0.18</version>
+			<scope>test</scope>
+		</dependency>
+		<dependency>
 			<groupId>org.mongodb</groupId>
 			<artifactId>mongo-java-driver</artifactId>
 			<version>3.12.2</version>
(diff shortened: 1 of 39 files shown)


dbluhm:
compile-testing is Apache 2.0. Is that compatible with Eclipse's requirements?

dbluhm:
For reference: https://github.com/google/compile-testing

jayjaybillings:
Thanks. Yes, that's fine.

dbluhm:
@jayjaybillings this one is ready for review. I opted to leave off tests for `@Persisted` and `@DataFieldJson` since I anticipate we'll need to make significant changes to those to bring them more in line with the new setup anyways.

jayjaybillings:
Approved and accepted. Thanks!

No need for CQ here plus @dbluhm is an employee of a member company.

PR #425, opened 2020-06-25
Notifications in Commands

osbornjd:
## Overview

This PR implements a notification scheme for commands. There are two notification options - via an POST call or via an email. 

### Design

The design is with an interface `ICommandUpdateHandler` which defines the necessary implementation for concrete implementations. The two implementations at the moment are `HTTPCommandUpdateHandler` and `EmailUpdateHandler`. Additional implementations could be added, if desired.

The HTTP notification takes a message and a URL and creates a POST call with the `httpclient` package from Apache. At the moment, is up to the user to properly format the POST call for their specific needs. There is nothing fancy about the call (e.g. with html or something else), it is a string and that string is what is sent to the POST call.

The `EmailUpdateHandler` takes a user provided email address, password, and email host and sends an email from that address to that address (with a specified message). This was the easiest conceivable way to implement this so that email passwords are kept private to the user of Commands. 

The actual `Command` is given an optional member variable which corresponds to the updater. If the Command has a non-null `ICommandUpdateHandler`, then upon the job "finishing" (in whatever state that may be), the update will be posted via the given UpdateHandler. 

### Note on Tests

The `HTTPHandlerTest` functions as is. The `EmailHandlerTest` requires some additional information, as is specified in the README. To pass, it requires a text file that is similar to the credential file for the `RemoteCommand` tests that contains the user's email address, password, and email host.


osbornjd:
I'll leave this up for the day, and if there are no objections will merge it tomorrow.

jayjaybillings:
Why do you need a password to send the email?

jayjaybillings:
Is it because you are using a corporate email server and not one running on localhost?

osbornjd:
The email client requires a password to authenticate - in principle if your email address didn't have a password and was open for anyone to use then it wouldn't need a password to authenticate. Yes, I was testing with a dummy gmail account.

jayjaybillings:
Can you test it running your own email server on localhost? I would like to explore a password-less option.

osbornjd:
> We need to review this class (and possibly others) to understand the password lifecycle. In principle, we should not store passwords as unencrypted member variables. (I acknowledge that this is a bigger issue than this PR, but needed to note it.)

 I agree. In `RemoteCommands` the password is never stored as it is just continuously passed through functions until it is handed off to Mina for verification (or an ssh key is used, which is the best option). I can look into using the `TxtFileConnectionAuthorizationHandler` to do the authorization, which doesn't currently store the password in a string.

> With respect to the ice-email-creds file, can this go in user.home instead of /tmp?

We could, but I would argue that we should leave it in `/tmp` because this is where the dummy remote credential file needs to go too. This was initially setup by Caleb in `/tmp` which is why I put it there. I think that both credential files should be in the same place for consistency, so I could do this but it would require changing all of the paths that use these files.

> Is CommandFactoryUpdaterTest actually checking for the receipt of an update or just the exit status?

It just checks for the exit status. However, the source code in `Command` that posts the update is the exact same as the `HTTPHandlerTest` and `EmailHandlerTest` source code, so as long as those pass `CommandFactoryUpdaterTest` will be successful. I could alternatively have `Command` throw an exception [here](https://github.com/eclipse/ice/blob/03d3063286760fdc84f94c713a42e2a642e6a682/org.eclipse.ice.commands/src/main/java/org/eclipse/ice/commands/Command.java#L120) which the test could check for. However, in my opinion the job shouldn't be marked as "failed" if an update can't be posted (since in principal the job could be successful, just the update post might not be), which is why I initially put just a warn log statement in this exception.

> EmailHandlerTest.testEmailNotificationPostUpdate() should explicitly catch exceptions and call fail() if it does around line 61.
> Same with HTTPHandlerTest around line 43.

Okay, changed.

osbornjd:
@jayjaybillings I've responded to/updated all your comments/requests. Let me know if you have additional comments.

jayjaybillings:
A few thoughts...
1. With respect to /tmp, that is really the wrong place to do this because it doesn't exist on all operating systems and it is hard coded. Although it a lot of work, I would greatly prefer that all of these be switched to user.home.
2. I made a note about the pipedream.net URL. If it is going to change at some point, ought we consider a config file some where?

We used to do all of these things in user.home/ICETests. It was really nice because you could run the tests from any OS and we had over fifty different configuration and test files in it.

osbornjd:
Okay, I'll put a ticket in to change all of these directories and address it in a later PR

jayjaybillings:
Works for me. Are you ready for this to be merged?

osbornjd:
Yes, I can go ahead and merge it.

jayjaybillings:
@osbornjd Following up on our discussion last week, is there any reason why this cannot be merged into jay/MarkIII and then into next? I've tagged next with a 3.x-Jsch tag to indicate that it is the last version with Jsch in it.

osbornjd:
Nope, I can create a PR for eclipse/jayMarkIIIMina into jay/MarkIII and then let you take over with any other tags you might want to create before merging into next.

jayjaybillings:
Make it so!

PR #426, opened 2020-06-29
Rearrange dev.annotations to add proxytest module

dbluhm:
Opening these changes up for comment. This change facilitates testing of generated DataElements using a separate maven module as a proxy tester. The proxy is necessary to test generated DataElements as Maven projects cannot use themselves as an annotation processor, as far as I have been able to discover.

Since the diff is a little hectic, here's an overview of the changes:
- Create `org.eclipse.ice.dev` project
- Move `org.eclipse.ice.dev.annotations` to child project of `org.eclipse.ice.dev`
- Create `org.eclipse.ice.dev.annotations.proxytest` child project (sibling to `dev.annotations`)
- Move what were essentially the proxy tests from `org.eclipse.ice.renderer` to `org.eclipse.ice.dev.annotations.proxytest`

Very open to feedback on naming of projects and the approach


jayjaybillings:
Does it build with something like the following?
```
mvn clean verify
```

dbluhm:
@jayjaybillings Yep, running `mvn clean verify` in the parent will compile and execute tests for both of the modules and running it from either module will, naturally, compile and execute tests for that module, assuming the dependencies are available through a `mvn install` first. That `mvn install` step isn't necessary when run from the parent.

jayjaybillings:
How well do they load up in Eclipse? Nested project support exists in Eclipse, but was a little dodgy the last time I tried it.

dbluhm:
The initial change is a little messy (at least the way I interact with git -- maybe doing it through eclipse is more graceful?). I had to remove the `org.eclipse.ice.dev.annotations` project from my project list but then adding `org.eclipse.ice.dev` was fairly smooth. After selecting the `org.eclipse.ice.dev` directory as a new project to import, it detected the nested projects and gave the option to load those as well with the end result of all three projects (`dev`, `dev.annotations`, and `dev.annotations.proxytest`) correctly imported in one action.

dbluhm:
If we don't have any major concerns with these changes, I'll go ahead and remove the WIP status on it

jayjaybillings:
I'm good with it.

jayjaybillings:
So is this ready to merge?

dbluhm:
I have no further changes planned so yep

jayjaybillings:
Accepted. Thanks!

@dbluhm is an employee of UT, a member company, and not CQ is required per the IP workflow.

PR #429, opened 2020-07-06
Update jay/MarkIII

osbornjd:
This PR updates the jay/MarkIII branch to all of the features that have been added to Commands in the past ~6 months. These features were put into the jay/MarkIIIMina branch to avoid conflicts with other code using Commands that was dependent on JSch, since one of the features introduced was to change to sshd mina dependency.


jayjaybillings:
Perfect. Thanks. BTW - I'm going to delete the jay/MarkIIIMina branch since the history is now in jay/MarkIII.

osbornjd:
Sounds great

PR #431, opened 2020-07-13
Fix proxytest pom setup

dbluhm:
A few typos were causing some unexpected behavior that I didn't initially catch due to a polluted environment (the digital kind).


jayjaybillings:
Accepted. Thanks!

This contribution is well below the CQ requirement and Daniel is an employee of a member company, so no issues accepting it.

PR #432, opened 2020-07-14
Support directly (de)serializing DataElements with Jackson

dbluhm:
This PR shuffles around our serialization logic to enable directly serializing or deserializing DataElements using Jackson. This will hopefully enable better integration of DataElements with other frameworks such as Drop Wizard.

As a summary of the changes, rather than directly interacting with `JsonNode`s, DataElements are now annotated to inform Jackson how to serialize it. DataElements are now also generated with a constructor annotated with `@JsonCreator` to mark it as the constructor to use for deserialization.

### DIFF
--- a/org.eclipse.ice.dev/org.eclipse.ice.dev.annotations/src/main/java/org/eclipse/ice/dev/annotations/processors/Field.java
+++ b/org.eclipse.ice.dev/org.eclipse.ice.dev.annotations/src/main/java/org/eclipse/ice/dev/annotations/processors/Field.java
@@ -126,7 +126,7 @@ public class Field {
 	}
 
 	/**
-	 * Return this Fields name ready for use in a method.
+	 * Return this Fields name ready for use in a method name.
 	 * @return capitalized name
 	 */
 	@JsonIgnore
@@ -136,6 +136,10 @@ public class Field {
 
 	/**
 	 * Return the appropriate getter method name for this field.
+	 *
+	 * Due to the use of the Lombok {@code @Data} annotatation on DataElements, by
+	 * Lombok convention, Getters for fields of type {@code boolean} use "is"
+	 * instead of "get".
 	 * @return getter method name
 	 */
 	@JsonIgnore
@@ -151,6 +155,9 @@ public class Field {
 
 	/**
 	 * Return whether this field has a getter, directly or via one of its aliases.
+	 *
+	 * This method is separate from {@code getAnyGetter()} despite being very
+	 * similar for ease of use in velocity templates.
 	 * @return true if getter present, false otherwise
 	 */
 	@JsonIgnore
@@ -159,8 +166,9 @@ public class Field {
 	}
 
 	/**
-	 * Return the name of any valid getter for this field.
-	 * @return getter name, null if none
+	 * Return the name of any valid getter for this field, either a direct getter
+	 * for the field or one of its aliases.
+	 * @return getter name, null if none present
 	 */
 	@JsonIgnore
 	public String getAnyGetter() {
@@ -229,8 +237,8 @@ public class Field {
 		}
 
 		/**
-		 * Take a raw string value and pass through without manipulating for use as
-		 * type.
+		 * Set type to string. Attempts to determine the type to mark whether it
+		 * is primitive or not.
 		 * @param type String representation of type of this Field
 		 * @return this
 		 */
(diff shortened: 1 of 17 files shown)


dbluhm:
Also noteworthy: the new arrangement actually simplified some of the interactions of DataElements with the PersistenceHandler, resulting in overall cleaner DataElement implementations.

org.eclipse.ice.dev/org.eclipse.ice.dev.annotations/src/main/java/org/eclipse/ice/dev/annotations/processors/Field.java
```diff
@@ -134,6 +134,50 @@ public String getNameForMethod() {
...
+		}
+		return prefix + getNameForMethod();
+	}
+
+	/**
+	 * Return whether this field has a getter, directly or via one of its aliases.
+	 * @return true if getter present, false otherwise
+	 */
+	@JsonIgnore
+	public boolean hasGetter() {
```
osbornjd, line 157: What is the reason for having this function only call another function? Are both used in different ways? I guess my question is, could they be combined into a single function, and if not why?

org.eclipse.ice.dev/org.eclipse.ice.dev.annotations/src/main/resources/templates/DataElement.vm
```diff
@@ -110,6 +124,19 @@ public class ${class} implements ${interface}, Serializable {
 		}
 	}
 
+	/**
+	 * Copy the contents of another element into this element.
+	 * @param element the element to copy
+	 */
+	@Override
+	public void copy(@NonNull $interface element) {
```
osbornjd, line 132: I'm guessing the `@NonNull` flag is a check that handles the null exceptions you previously had in copy?

org.eclipse.ice.dev/org.eclipse.ice.dev.annotations/src/main/java/org/eclipse/ice/dev/annotations/processors/Field.java
```diff
@@ -134,6 +134,50 @@ public String getNameForMethod() {
...
 	}
 
+	/**
+	 * Return the appropriate getter method name for this field.
+	 * @return getter method name
+	 */
+	@JsonIgnore
+	public String getGetterName() {
+		String prefix = null;
+		if (type != null && type.equals("boolean")) {
```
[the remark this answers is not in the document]
dbluhm: This is a convention we inherit from use of Lombok. I'll add that to the doc comment to make sure that tidbit of information isn't lost.

org.eclipse.ice.dev/org.eclipse.ice.dev.annotations/src/main/java/org/eclipse/ice/dev/annotations/processors/Field.java
dbluhm: Good question. This method is intended for use from the velocity template where holding on to values is possible but not necessarily very clean. Having the separate boolean returning method reduces the amount of logic required in the template and, if logic changes in the future, that change would be transparent to the template. As an example of where this is used in the template, here is the copy method template:
```
	/**
	 * Copy the contents of another element into this element.
	 * @param element the element to copy
	 */
	@Override
	public void copy(@NonNull $interface element) {
		#foreach($field in $fields)
			#if($field.hasGetter())
		this.${field.Name} = element.${field.AnyGetter}();
			#end
		#end
	}
```

Without the separate `hasGetter()`:
```
	/**
	 * Copy the contents of another element into this element.
	 * @param element the element to copy
	 */
	@Override
	public void copy(@NonNull $interface element) {
		#foreach($field in $fields)
			#set($getter = ${field.AnyGetter})
			#if($getter)
		this.${field.Name} = element.$getter();
			#end
		#end
	}
```
Again, possible but, in my opinion at least (and I'm not super set on this), having the boolean `hasGetter()` seems better.

org.eclipse.ice.dev/org.eclipse.ice.dev.annotations/src/main/resources/templates/DataElement.vm
dbluhm: It's not too obvious from the template but I realized that some of the previous checks in the copy constructor were pointless, i.e. I had an `instanceof` check that literally amounted to `if (generatedDataElementImplementationInstance instanceof GeneratedDataElementImplementation) ...`. So removing that, the only thing left was a null check that threw a generic exception which I also realized was no different from the null pointer exception that is thrown by the [`@NonNull` annotation](https://projectlombok.org/features/NonNull).

jayjaybillings:
@osbornjd Any remaining concerns seeing @dbluhm's responses?

osbornjd:
None from me, I'll go ahead and merge it.

jayjaybillings:
@osbornjd Thanks!

PR #430, opened 2020-07-10
POJO Generation from JSON Utility

dbluhm:
This PR introduces a CLI utility for generating POJOs from JSON using the framework we've built out for the annotations in `org.eclipse.ice.dev.annotations`.

PR #434, opened 2020-07-16
W.I.P. DataModel Annotation

walshmm:
Still needs a lot of cleaning up, but here's what I have so far.


dbluhm:
Pointing out a bit of an overlap between this PR and #430. It seems we both needed to make the filling out and producing a file from a template accessible from outside the DataElementProcessor. For further assessment, this commit is the most relevant: a7b07e01e83cb0d66a6d4799bf05c447a5c0a5df

PR #441, opened 2020-07-23
Project generator archetype bug fixes

gecage952:
Removed the parent dependencies from the archetype poms. You can now build them each individually without needing the overall project jar in your local m2 repo.


jayjaybillings:
Approved. Thanks.

Accepted as below the CQ limit, plus @gecage952 is an employee of a member company.

PR #442, opened 2020-07-24
Fix references to IDataElement, etc. after move

dbluhm:
Signed-off-by: Daniel Bluhm <bluhmdj@ornl.gov>


jayjaybillings:
Thanks for catching these. I guess the templates make sense, but I'm not sure why Eclipse missed the .java files. I found one that it missed and fixed it too. Oh well. Good work!

Accepted. It's below the CQ limit and @dbluhm is an employee of a member company.

PR #444, opened 2020-07-28
Integration tests and test utilities

dbluhm:
After some discussion with @jayjaybillings on the best approach to testing our Persistence Handlers, it became clear we needed integration tests for them. This PR implements those integration tests.

To have these tests run at a later stage from unit tests, I created a new package for them. Additionally, as these integration tests will have environment prerequisites and it would probably be annoying to have them nested alongside the unit tests if you don't have everything set up, I opted to place it in a new top-level project `org.eclipse.ice.tests.integration`. We can use this package for any other needed integration tests.

CC: @osbornjd


dbluhm:
> Why are so many tests using JsonIgnoreUnknown?

It's a little deceiving but the added `JsonIgnoreUnkown` annotation are on the patterns used for validating the generated data element implementation. Since I added the annotation to the data element template, each of the patterns had to be updated to reflect that (annotations are considered part of the class definition in the AST and therefore must be present to match).

jayjaybillings:
Thanks for the extra info.

How soon do you want these tests merged into next? Can we wait on @osbornjd's changes while you pick something else up or is this blocking you?

osbornjd:
That is why I asked whether or not I was going to delay this PR yesterday, as I probably won't be super fast about implementing my changes (and don't want to hold things up if it is important to merge it).

dbluhm:
I think there's a relatively clean split between the tasks so accepting this one (once I have the test utilities ironed out, I might set this to WIP until I am satisfied with those) with a PR later from Joe seems like a fair idea.

jayjaybillings:
I'll take it separately when you're ready then.

jayjaybillings:
You've clearly learned to predict what I will complain about. :-P Great work! Accepted.

Accepted since @dbluhm is an employee of a member company and all checks pass.

PR #449, opened 2020-08-03
Annotation processor logging

dbluhm:
This PR contains an experiment in propagating warnings and errors logged by libraries utilized by our annotation processors through the `javax.annotation.processing.Messager`. This ensures errors and warnings are not missed during compilation of clients and also makes it easier to catch these errors in our own testing.

### DIFF
--- a/org.eclipse.ice.dev/org.eclipse.ice.dev.annotations/pom.xml
+++ b/org.eclipse.ice.dev/org.eclipse.ice.dev.annotations/pom.xml
@@ -102,6 +102,11 @@
 			<version>3.0.0-SNAPSHOT</version>
 			<scope>compile</scope>
 		</dependency>
+		<dependency>
+			<groupId>com.cosium.logging</groupId>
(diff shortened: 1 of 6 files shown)


dbluhm:
For reference, here is the logger annotation library used: https://github.com/Cosium/annotation-processor-logger

jayjaybillings:
In principle that dependency should be OK. The license is one that we can accept. I'll review the code further.

dbluhm:
> Why does VelocityProperties need to change?

Strictly speaking it doesn't need to change as much as I changed it to get to passing tests again. Since I was looking at it again, I realized I was just statically defining properties so might as well put it into a more cleanly version controlled properties file. I'm fine with reverting that, though, and having a cleaner change set here.

dbluhm:
To be more explicit, there are some things that did need to change in VelocityProperties but just the property names themselves. Everything else could have remained the same.

jayjaybillings:
We're still WIP on this right?

PR #454, opened 2020-08-06
Automatically import types

### LINKED ISSUE #450: @DataElement has trouble with enums
This is probably low priority, but here's an example of using an @DataElement annotated class with an enumerated value for a member.

```java
@DataElement(name="TaskStateData")
public class TaskStateDataSpec {

	/**
	 * The state of the task
	 */
	@DataField.Default(value = "org.eclipse.ice.tasks.TaskState.INITIALIZED")
	@DataField
	private TaskState taskState;
	
}
```

Note that the full `org.eclipse.ice.tasks.TaskState.INITIALIZED` declaration is required because the `value` tag is a pass through and `import org.eclipse.ice.tasks.TaskState;` is not included.

dbluhm:
This PR adds automatic importing of DataField types when necessary to classes generated through DataElement. This results in cleaner generated classes (look more like something you'd actually write by hand) and also makes specifying defaults more straightforward (for instance, when the invoking a class that is imported, you can use the string `new MyClass()` instead of `new com.example.MyClass()`).

As a side effect of these changes, I removed generics from JavascriptValidator. This significantly simplified the resolution of imports as this previously required putting velocity variables that would later be evaluated and interpolated into the JavascriptValidator's type. The character used by velocity for string interpolation is `$` which collides with java's internal representation of inner classes (i.e. `java.util.Map.Entry` can be expressed as `java.util.Map$Entry`) which is problematic for parsing out types to import.

Fixes #450.


dbluhm:
Should also note that:
- This PR will handle name collisions between imports and revert to the fully qualified name of the elements in question
- All types are maximally specified meaning that inner classes (like `Map.Entry`) are not accessed through their parent and are imported directly (`Map.Entry` data fields will cause an import of `java.util.Map.Entry` and `Entry` will be used in all cases where the field type is used).

PR #455, opened 2020-08-06
Fix VelocityProperties and Annotation Processor Logging

dbluhm:
This PR fixes two issues:
- VelocityProperties was set to load from a file when it should have been loading from a resource
- Annotation processor logging the way I set it up would conflict with any logger implementations put on dependent projects. I reverted the main `DataElementProcessor` to extend the usual `AbstractProcessor` and wrote a quick wrapper around it for use in testing that adds in the logging functionality. With that in place, the annotation-processor-logging library is a test only dependency, solving our logger conflict problems.


dbluhm:
And, while I'm fixing things, also fixed the group ID used by the persistence integration tests; @gecage952's PR correcting the group IDs was merged at about the same time as the persistence integration test PR and that last instance persisted (lol) as a result.

PR #456, opened 2020-08-07
Fix builds, revert JavascriptValidator generics removal

dbluhm:
I added `org.eclipse.ice.data` to the build and discovered there were more issues with JavascriptValidator than I originally thought. I reverted the removal of the generics and fixed my type lookup mechanism to accommodate passing through values to be interpolated by velocity. I'm not particularly fond of that need but I think it's better to just fix the issues and address the bigger issue down the road.

### DIFF
--- a/org.eclipse.ice.build/pom.xml
+++ b/org.eclipse.ice.build/pom.xml
@@ -1,31 +1,30 @@
 <?xml version="1.0" encoding="UTF-8"?>
 
 <project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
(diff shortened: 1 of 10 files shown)


org.eclipse.ice.build/pom.xml
```diff
@@ -1,31 +1,30 @@
...
-  	<groupId>org.eclipse.ice</groupId>
-  	<artifactId>org.eclipse.ice.parent</artifactId>
-  	<version>2.2.1-SNAPSHOT</version>
-  	<relativePath>../org.eclipse.ice.parent/pom.xml</relativePath>
-  </parent>
-  <groupId>org.eclipse.ice</groupId>
-  <artifactId>org.eclipse.ice.build</artifactId>
-  <version>1.0-SNAPSHOT</version>
-  <packaging>pom</packaging>
+	<parent>
```
jayjaybillings, line 7: This versioning looks wrong. Everything should be 3.0. Also, I'm not sure why we are using the old parent.

PR #457, opened 2020-08-07
New Tasks infrastructure to replace Item

jayjaybillings:
The Tasks library will replace the Item from 2.0. Documentation is included. Test coverage is ~84.0%


org.eclipse.ice.tasks/src/main/java/org/eclipse/ice/tasks/Task.java
```diff
@@ -0,0 +1,270 @@
...
+			builder.configureTransitions().withExternal().source(TaskState.READY).target(TaskState.EXECUTING)
+					.event(TaskTransitionEvents.EXECUTION_TRIGGERED)
+					.action(new ExecutingEventAction<>(stateData.get()));
+			// ...and a state action. This action requires all of the data.
+			builder.configureStates().withStates().state(TaskState.EXECUTING, smExecutingAction);
+
+			// If the execution finished as expected, there is a transition event for that
+			// to finalize the state.
+			builder.configureTransitions().withExternal().source(TaskState.EXECUTING).target(TaskState.FINISHED)
+					.event(TaskTransitionEvents.EXECUTION_FINISHED).action(new FinishedEventAction<>(stateData.get()));
```
dbluhm, line 260: In the event that the execution did not finish as expected, should it enter the `TaskState.FAILED` state? Does that transition need to be defined here also?

jayjaybillings:
A quick note here: For some reason the validator is picking up my old ornl.gov address even though it isn't showing in the Git history. I suspect this is a bug and we can ignore it. If it becomes a problem, I can raise it with the EMO.

org.eclipse.ice.tasks/src/main/java/org/eclipse/ice/tasks/Task.java
jayjaybillings: With respect to differences...
1. See the code. This uses a new state machine library instead of the one I wrote from scratch. Also much cleaner.
2. See my thesis. The workflow model matches that with some corrections.

jayjaybillings: With respect to entering TaskState.FAILED, yes it should. However, that is a state action and what you are looking at here is a transition action. So the movement to TaskState.Failed has to be triggered from within the Action itself and isn't listed here.

PR #461, opened 2020-08-21
Dependency scraper

dbluhm:
This PR adds a maven plugin that can be to retrieve files from dependencies and copy them to another location in the current project. This is primarily intended for use in Vaadin projects that use Data Elements from dependencies and being able to pull TypeScript representations of those data elements and other auto-generated TypeScript classes into the projects frontend sources.

This PR formerly included basic TypeScript generation but I removed it with the intent of keeping the change set limited for this PR.


org.eclipse.ice.dev/org.eclipse.ice.dev.dependencyscraper/src/main/java/org/org/eclipse/ice/dev/dependencyscraper/DependencyScraper.java
```diff
@@ -0,0 +1,227 @@
...
+						"File %s already exists and clobber is not set; skipping.",
+						target.getName()
+					));
+				}
+			} else {
+				FileUtils.copyInputStreamToFile(
+					jar.getInputStream(fileInJar),
+					target
+				);
+			}
```
jayjaybillings, line 190: I would prefer to break this into at least two functions because there are four layers of branching.

org.eclipse.ice.dev/org.eclipse.ice.dev.dependencyscraper/src/main/java/org/org/eclipse/ice/dev/dependencyscraper/DependencyScraper.java
```diff
@@ -0,0 +1,227 @@
...
+				for (ZipEntry file : toCopy) {
+					getLog().info(String.format("Copying %s", file.getName()));
+					copyFileInJarToOutput(jarFile, file);
+				}
+			} catch (IOException e) {
+				throw new MojoFailureException(
+					String.format("Failed to open jar file %s!", jar.toString())
+				);
+			}
+		}
```
jayjaybillings, line 225: More documentation would be nice, even if only a few lines. All of the big functional blocks, like 205-207 and 211-214 should have at least a few lines describing what they are doing because functional code is inherently confusing. Personally I would put one line per filter in addition a one line initial description.
```java
// Get the set of files to copy
Set<ZipEntry> toCopy = jarFile.stream()
   // No directories
   .filter(file -> !file.isDirectory())
   // Is in the source directory
   .filter(this::startsWithSourceDirectory)
   // Matches the inclusion filter for this type
  .filter(this::shouldInclude).collect(Collectors.toSet());
```

org.eclipse.ice.dev/org.eclipse.ice.dev.dependencyscraper/src/test/java/org/org/eclipse/ice/dev/dependencyscraper/DependencyScraperTest.java
```diff
@@ -0,0 +1,193 @@
...
+import org.junit.Test;
+import java.io.File;
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.attribute.BasicFileAttributes;
+import java.util.List;
+import java.util.Set;
+
+public class DependencyScraperTest
```
jayjaybillings, line 18: Documentation please

org.eclipse.ice.dev/org.eclipse.ice.dev.dependencyscraper/src/test/java/org/org/eclipse/ice/dev/dependencyscraper/DependencyScraperTest.java
```diff
@@ -0,0 +1,193 @@
...
+{
+	private static final Path PROJECT = Path.of("target/test-classes/project-to-test");
+	private static final Path OUTPUT = Path.of("target/test-classes/project-to-test/test");
+
+	@Rule
```
jayjaybillings, line 23: What's the rule do? (i.e. - Docs please!)

jayjaybillings (changes requested):
What's the src/it folder for? Are we now using Groovy as a dependency?

Your test looks comprehensive. What's your coverage level?

dbluhm:
> What's the src/it folder for? Are we now using Groovy as a dependency?

Ah, woops, that was included in the archetype I used to generate the base project. Fortunately, the plugin is simple enough that I don't think we'll require integration tests so I will delete that and the associated pieces. Thanks for catching that.

> Your test looks comprehensive. What's your coverage level?

88.5% which is basically everything but a couple of `IOException` handlers. Initially I wasn't sure I'd be able to get it to hit those handlers in testing but looking at it again I think I could probably hit at least one scenario that would throw an IO Exception.

dbluhm:
Addressed feedback and increased coverage to 98% :slightly_smiling_face:

jayjaybillings:
Nice work on the coverage! :) Is there any reason that this is not merged other than me? I think you addressed my questions, correct?

dbluhm:
> Nice work on the coverage! :) Is there any reason that this is not merged other than me? I think you addressed my questions, correct?

I think that's everything! Just waiting to make sure I'd adequately addressed your comments before merging

jayjaybillings (approved):
Nice work.

PR #463, opened 2020-08-28
Update READMEs

osbornjd:
This PR updates the various READMEs related to Commands that are now in different places after the integration test migration. It also creates a README for the integration test area.

Signed-off-by: Joe Osborn <osbornjd@ornl.gov>


osbornjd:
Merging this since it doesn't actually change any functionality.

PR #468, opened 2020-09-11
Typescript generation

dbluhm:
This introduces rudimentary typescript model generation from data elements. When combined with the [Maven Resources Plugin](https://maven.apache.org/plugins/maven-resources-plugin/examples/copy-resources.html) and the dependency scraper #461(when it is desired to retrieve this typescript model from a dependency jar), these generated models can be moved to the frontend directory of Vaadin projects and used in Lit Element Templates.


dbluhm:
Included in the PR is an example of using the Maven Resources Plugin in org.eclipse.ice.renderer to move the generated TypeScript file for the Person data element to a location that can subsequently be scrapped using the depedency scraper.

dbluhm:
Also added an example to `rse-renderer-client-test` of one way we can load in it's own auto-generated typescript models. One interesting thing I discovered is that the auto-generated `.ts` files _can_ be loaded without being moved around but the challenge with leaving them where they are is that the path to import them would be pretty ugly. I experimented with modifying the `webpack.config.js` file to add aliases to DataElement so the deeply nested, generated TypeScript would compile correctly and this approach of adding aliases to the resolution of webpack may be a better way to handle use of generated TypeScript in the future.

dbluhm:
To add to my last comment, we can likely follow the same approach that Vaadin uses and make a `webpack.ice.js` that is required and merged with the `webpack.config.js`. `webpack.ice.js` would get generated on each compile by a plugin that aggregates a list of all the auto-generated TypeScript. It is notable that the dependency scraper is still definitely needed for pulling typescript out of the dependencies but using this `webpack.ice.js`, rather than putting the scrapped typescript directly into the `frontend/src` directory, we could leave it somewhere in `target`. I think that will prevent any accidental attempts to modify an auto-generated file.

dbluhm:
Figured out how to get nice imports without moving things around from the generated source directory to the frontend directory in a private branch. I'll likely update this PR with those changes. They ultimately come out as this in the `webpack.config.js`:

```js
const GENERATED_PREFIX = '@Generated/';

/**
 * Recursive file matching with regex.
 */
function fileMatch(dir, regex) {
  let files = fs.readdirSync(dir, {withFileTypes: true});
  return files.reduce((acc, file) => {
    if (file.isDirectory()) {
      acc.push(...fileMatch(`${dir}/${file.name}`, regex));
    } else if (file.isFile() && file.name.match(regex)) {
      acc.push(`${dir}/${file.name}`);
    }
    return acc;
  }, []);
}

/**
 * Generated aliases for auto-generated typescript files in this project.
 */
function aliasGeneratedFiles() {
  let generated = fileMatch('target/generated-sources', /^[^.]+(?!\.d).ts$/);
  let aliases = generated.reduce((acc, file) => {
    acc[`${GENERATED_PREFIX}${path.basename(file, '.ts')}$`] = path.resolve(__dirname, `${file}`);
    return acc;
  }, {});
  console.warn('Generated aliases: ', aliases);
  return aliases;
}

module.exports = merge(flowDefaults, {
  resolve: {
    extensions: ['.ts', '.tsx', '.js', '.json'],
    alias: {
      DataElement$: path.resolve(__dirname, 'frontend/src/data-element.ts'),
      ...aliasGeneratedFiles()
    },
    modules: ['target/generated-sources', 'node_modules']
  },
  devtool: 'inline-source-map',
  module: {
    rules: [{
      // Include ts, tsx, js, and jsx files.
      test: /\.tsx?$/,
      exclude: /node_modules/,
      loader: 'ts-loader',
    }],
  }
});
```

We should be able to add this to the archetype to have it included in future projects

PR #471, opened 2020-09-29
Use Builder for JSON Deserialization

dbluhm:
I found that the previous `@JsonCreator` method had some flaws in practice. For instance, if your JSON you're deserializing from omits any of the default fields of DataElements, deserialization would fail due to `@NonNull` being used. Removing that, you get `null`s instead of the defaults declared on the DataElement.

Using the Builder form here, we get to use sparse JSON and have the defaults placed correctly when using sparse JSON.


jayjaybillings:
Is there any problem with merging this?

jayjaybillings (approved):
Approved and ready to merge when you're ready.

dbluhm:
> Is there any problem with merging this?

Don't think so -- tests are passing and we've done some vetting and everything seems good.

PR #473, opened 2020-10-02
Fix running with Java versions > 11 failing tests and windows compat fixes

### LINKED ISSUE #472: Supported source version 'RELEASE_11' from annotation processor 'org.eclipse.ice.tests.dev.annotations.processors.LoggingDataElementProcessor' less than -source '14'
I encounter this error - `Supported source version 'RELEASE_11' from annotation processor 'org.eclipse.ice.tests.dev.annotations.processors.LoggingDataElementProcessor' less than -source '14'` - when I try to build on win64 with Java 14.

It is caused by using `@SupportedSourceVersion(SourceVersion.Release_11)` on the annotation processors. 

Do we actually need this? Is there any reason why we would want to limit the latest version of JDK that we could use? I would propose that we remove these.

@dbluhm What do you think? If you agree I'll push the changes.

dbluhm:
DataElementProcessor tests loosened to not consider warnings a failure
condition. Removed LoggingDataElementProcessor in favor of just having a
standard logger during testing; leave it up to tester to determine if
warnings are serious or not.

Closes #472


jayjaybillings:
Merged. I also added some source, target, and release build restrictions.

PR #476, opened 2020-10-23
Use metadata classes in writers directly

dbluhm:
Instead of having separate builders to transfer the exact same metadata. Also simplified PojoFromJson by removing PojoOutline which is now redundant given that DataElementMetadata can be used directly in its place.


jayjaybillings (approved):
This is great! I strongly prefer it to the builder-based approach. Did you update the examples in the README?

dbluhm:
When I was writing out the README was actually when I had my "Oh, this could be a lot simpler" moment lol so I followed this pattern in the example for the README.

PR #483, opened 2020-10-29
Adds unit tests and delete method to BATS

marshallmcdonnell:
Work includes:
  - Added delete method to BATS and an integration test for the method
  - Added unit tests for getter / setter methods of DataSet
  - Switched to spaces instead of tabs in files

**DISCLAIMER:** 
The switching of spaces to tabs really kills the view of changes over all the commits.
Probably need to look at individual commits to see the actual changes.
Apologies to reviewers.


jayjaybillings (approved):
Looks good, thanks. All checks pass and you are an employee of a member company, so accepted.

PR #486, opened 2020-10-29
JsonSchemaConverter File name tests & remove org.eclipse.ice.data.neutrons

gecage952:
This pr contains the file name tests for the json schema converter. I also realized that when merging previous changes, on accident I removed a couple of lines in PojoFromJson, and I added them back here as they should be. I also went ahead and deleted org.eclipse.ice.data.neutrons, since that was kind of a demo for Oak Ridge stuff, but I don't think there's much merit to having it here at the moment. I did keep some back ups of it in case we really need it for some reason.


gecage952:
I'll just at @dbluhm and @jayjaybillings  here.

dbluhm (approved):
Looks good, thanks for getting those tests in!

PR #491, opened 2020-11-12
Create top-level POM for ICE + BATS refactoring

### LINKED ISSUE #488: Make a top-level ICE pom.xml from org.eclipse.ice.build
From the off-topic discussions in a closed issue, had discussions to potentially move the pom.xml file in [`org.eclipse.ice.build` pom file](https://github.com/eclipse/ice/blob/next/org.eclipse.ice.build/pom.xml) to the top-level of ICE:

Key conversation links:
 - Suggestion part 1: https://github.com/eclipse/ice/issues/475#issuecomment-717271343
 - Suggestion part 2: https://github.com/eclipse/ice/issues/475#issuecomment-717988044
 - Response: https://github.com/eclipse/ice/issues/475#issuecomment-718000006

Would like to move this to the top-level so ICE has one "main pom.xml" file.
Will help using JitPack in other projects that have ICE sub-projects as dependencies ([example for multi-module repo](https://github.com/jitpack/maven-modular))

marshallmcdonnell:
Fixes #488 

*Work includes:*
 - Moving the `org.eclipse.ice.build` pom.xml to the top-level of the project
 - Removing the `org.eclipse.ice.build` package entirely
 - Modifying BATS so you no longer need to have the `docker:build` if you do `mvn clean docker:build verify` (now just need `mvn clean verify`
 - Removed BATS sub-module from the top-level pom.xml `default` profile
 - Added a `full` profile to top-level pom.xml that includes building ICE with BATS (requires Docker)

*To test:*
1) Ensure can build ICE at the top-level using `mvn clean verify` and/or `mvn clean install`
2) When doing this, ensure BATS is NOT built
3) Now, with Docker installed, go to `org.eclipse.ice.bats` module and run `mvn clean verify`. Should see that a Fuseki docker image is built in the `pre-integration-test` phase and the integration tests run.
4) Go back to the top-level of ICE and now run the `full` profile so we including building BATS. Run `mvn clean verify -P full` and/or `mvn clean install -P full`. Ensure BATS is built.


dbluhm (approved):
Tested and working!

